PR #18, opened 2019-11-14
update `equalslice` functions to use constant time comparison

rjrbt:
This is a PR for a small refactor to `equalslice` functions to use a constant time comparison similar to that found in the `subtle` library.

This allows for two slices, a and b, to always return at the same response rate as a successful comparison.

### DIFF
--- a/bfv/utils.go
+++ b/bfv/utils.go
@@ -7,12 +7,11 @@ func equalslice(a, b []uint64) bool {
 		return false
 	}
 
+	v := true
 	for i := range a {
-		if a[i] != b[i] {
-			return false
-		}
+		v = v && (a[i] == b[i])
 	}
-	return true
+	return v
 }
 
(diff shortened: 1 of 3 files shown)


ChristianMct:
Hey! Thank you for pointing this out. Although these non-constant time functions are only used for test purposes, it won't hurt much having them secured.

rjrbt:
@ChristianMct - you know. I thought that too, but one instantance in bfv is used in the exported `Equals()` method. 
https://github.com/ldsec/lattigo/blob/12ff58e9e1eec789570641d6a547be2204708eea/bfv/params.go#L91

ChristianMct:
Good catch! One more reason to fix it. We can do that for the next release.

ChristianMct:
The issue was solved as a part of the extraction of a common `utils` package (commit bda3455f1428c69945332ff704b36a82ca94523f).

PR #20, opened 2019-11-22
Context becomes parameters

cgrigis:
WiP, to be discussed whether I am in the right direction. :)

Fixes c4dt/lattigo#1

### DIFF
--- a/bfv/bfv_benchmark_test.go
+++ b/bfv/bfv_benchmark_test.go
@@ -34,7 +34,7 @@ func Benchmark_BFV(b *testing.B) {
 		encryptorPk := NewEncryptorFromPk(pk, &params)
 		encryptorSk := NewEncryptorFromSk(sk, &params)
 
-		ringCtx := NewCiphertextRingContext(&params)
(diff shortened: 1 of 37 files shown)


ChristianMct:
This is the direction we want to take for these pre-computed values. Is there a specific reason why you think `EncoderContext` should be a public type and have a public constructor ? We didn't plan on exposing these, initially.

cgrigis:
> Is there a specific reason why you think `EncoderContext` should be a public type and have a public constructor ? We didn't plan on exposing these, initially.

Good point, no specific reason, making it private now.

cgrigis:
So, my understanding is that the same approach goes for `NewKeyGenerator()`, `newEncryptor()`, `NewDecryptor()`, `NewEvaluator()`, `NewPlaintext()` and `NewCiphertext()`, correct?

ChristianMct:
That's correct. Moreover, most of these contexts structs will have
common attributes. So this could be done elegantly using struct embedding.

ChristianMct:
Close this PR as now duplicated.

PR #24, opened 2019-12-12
change Encryptor to an interface

cgrigis:
This PR changes the `bfv.Encryptor` struct to an interface, adding a small hierarchy where `pk` and `sk` only exist in their respective implementation, and removing the "ugly" tests such as `if pk != nil {...} else if sk != nil {...}`.

This can also apply to `ckks.Encryptor`. For symmetry and abstraction, we can also change `Decryptor` to an interface (even though it only has one implementation), `Encoder`, etc.

What do you think?


ChristianMct:
Agreed! Much nicer and modular.

PR #41, opened 2020-03-27
use more require in tests

tharvik:
So, I've been updating some test files to use require instead of `t.{Error,Fatal}`. Also (and that was the initial drive for my PR), I've added some check in `ckks/marshaler.go:UnmarshalBinary` to ensure that the given bytearray was fully used.


ChristianMct:
These changes can be integrated in the v1.4.0 upcoming release.

ChristianMct:
Since conflicts with `v1.4.0` were small, I took care of them. Your commits are now part of the `de_v1.4.0` branch. Thanks for the contribution!

tharvik:
Happy to help :)

PR #47, opened 2020-05-15
[utils] Clean up PRNG interface

eliedz:
Adds a common interface for prng that encompasses a keyed prng using black2b and a rand prng that reads a key from crypto/rand


ring/prng.go
```diff
@@ -14,19 +14,25 @@ type CRPGenerator struct {
...
 	sum     []byte
 	masks   []uint64
+	seeded  bool
 }
 
 // NewCRPGenerator creates a new CRPGenerator, that will deterministically and securely generate uniform polynomials
 // in the domain of the input context using the hash function blake2b. The PRNG can be instantiated with a key on top
 // of the public seed. If no key is used, set key=nil.
-func NewCRPGenerator(key []byte, context *Context) *CRPGenerator {
+func NewCRPGenerator(seeded bool, key []byte, context *Context) *CRPGenerator {
```
ChristianMct, line 23: I would rather explicitly pass the PRNG to the generator when I need to have control over it:
```suggestion
func NewCRPGeneratorFromPRNG(prng utils.PRNG, key []byte, context *Context) *CRPGenerator {
```

and have a default constructor using the hash function:

```suggestion
func NewCRPGenerator(key []byte, context *Context) *CRPGenerator {
     return NewCRPGeneratorFromPRNG(utils.NewSeededPRNG(key))
}
``` 

Also above: if a random polynomial generator is not seeded, there shouldn't be any `key` passed. So I would see `key` as a parameter to the PRNG.

utils/prng.go
```diff
@@ -1,73 +1,80 @@
...
+	}
+}
+
+// seededPRNG is a structure storing the parameters used to securely and deterministically generate shared
 // sequences of random bytes among different parties using the hash function blake2b. Backward sequence
 // security (given the digest i, compute the digest i-1) is ensured by default, however forward sequence
-// security (given the digest i, compute the digest i+1) is only ensured if the prng is keyed.
-type prng struct {
+// security (given the digest i, compute the digest i+1) is only ensured if the seededPRNG is keyed.
+type SeededPRNG struct {
```
ChristianMct, line 27: Does this type needs to be public if it is hidden behind the PRNG interface ?

ChristianMct, line 27: Since we are implementing the PRNG with Blake2, could a better name be `blake2PRNG`

ring/prng.go
eliedz: Even if it is seeded, passing a key is optional for blake2b so it is not a differentiating parameter for PRNG. 
As for explicitely passing the PRNG to the generator, I agree it would be better so to allow reusing a PRNG but there would still need to be a boolean to identify if the PRNG being passed is seeded or not.

utils/prng.go
eliedz: I thought it would be better to generalize the naming in case there are other seeded prngs added

eliedz: It needs to be exposed as it implements some functions outside the interface that might be called

ChristianMct: So let's make that an interface ? 

```suggestion
type SeededPRNG interface {
	PRNG
	Seed(seed []byte)
	GetSeed() []byte
	SetClock(sum []byte, n uint64) error
}
```

ring/prng.go
ChristianMct: It seems that the only access to that parameter is to determine if the generator panics or not. We could resolve this by differentiating between two interfaces `SeededPolynomialGenerator` and `PolynomialGenerator`, as they are not used for the same purpose.

ChristianMct: Regarding the parameters, I'm confused between the `seed` and the `key`. If I remember well, all we want is to be able to produce common random polynomials given a common random `[]byte`. Do we need the two concepts (key and seed) or can we simplify ?

utils/prng.go
ChristianMct: Depending on the discussion of [this comment](https://github.com/ldsec/lattigo/pull/47#discussion_r430545437), we can change the interface name to `KeyedPRNG`. 

The `sum` variable is also not documented. Can we also simplify here ?

ring/ring_benchmark_test.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func BenchmarkRing(b *testing.B) {
-	b.Run("GenRingContext", benchGenRingContext)
+	//b.Run("GenRingContext", benchGenRingContext)
```
ChristianMct, line 11: Is there a reason for removing this benchmark ?

ChristianMct:
Following the clean up of `utils/prng.go`, we also need to clarify and optimize the `ring` package.

As of now, sampling of ring elements is done by calling methods on the `Context` that are defined in:

- `sampler.go` for uniform sampling in R_q
- `prng.go` for seeded uniform sampling  in R_q
- `ternarySampler.go` for uniform sampling in R_3
- `gaussianSampler.go` for gaussian sampling

All of these samplers use `rand.Read` and should now use `utils.PRNG`. The proposition is to extract three structs:

- `UniformSampler`
- `TernarySampler`
- `GaussianSampler`

Each struct would wrap a `utils.PRNG` given in its constructor and expose a `Read(p *ring.Poly)` method to sample a new value into `p`. Hence, the functionality of `ring.CRPGenerator` is covered by the `UniformSampler` initialized with a `utils.KeyedPRNG`. This will also enable to reduce the size of the `ring.Context`, since the sampling-exclusive fields can be moved to the sampler structs.

ring/ring_benchmark_test.go
eliedz: No, my bad

dckks/relinkey_gen.go
```diff
@@ -65,6 +71,11 @@ func (ekg *RKGProtocol) NewEphemeralKey(p float64) (ephemeralKey *ring.Poly) {
 func (ekg *RKGProtocol) GenShareRoundOne(u, sk *ring.Poly, crp []*ring.Poly, shareOut RKGShareRoundOne) {
 
 	contextQP := ekg.dckksContext.contextQP
+	prng, err := utils.NewPRNG()
+	if err != nil {
+		panic(err)
+	}
+	gaussianSampler := ring.NewGaussianSampler(prng, contextQP)
```
ChristianMct, line 78: The gaussian sampler could be pre-allocated as a field in the `RKGProtocol` type

dckks/publickey_gen.go
```diff
@@ -38,7 +39,13 @@ func (ckg *CKGProtocol) AllocateShares() CKGShare {
 // for the receiver protocol. Has no effect is the share was already generated.
 func (ckg *CKGProtocol) GenShare(sk *ring.Poly, crs *ring.Poly, shareOut CKGShare) {
 	contextQP := ckg.dckksContext.contextQP
-	contextQP.SampleGaussianNTTLvl(uint64(len(contextQP.Modulus)-1), shareOut, ckg.dckksContext.params.Sigma, uint64(6*ckg.dckksContext.params.Sigma))
+	prng, err := utils.NewPRNG()
+	if err != nil {
+		panic(err)
+	}
+	gaussianSampler := ring.NewGaussianSampler(prng, contextQP)
```
ChristianMct, line 46: The gaussian sampler can be pre-allocated in the `CKGProtocol` type.

dckks/public_refresh.go
```diff
@@ -61,6 +62,12 @@ func (refreshProtocol *RefreshProtocol) GenShares(sk *ring.Poly, levelStart, nPa
 		}
 	}
 
+	prng, err := utils.NewPRNG()
+	if err != nil {
+		panic(err)
+	}
+	gaussianSampler := ring.NewGaussianSampler(prng, context)
```
ChristianMct, line 69: The gaussian sampler can be allocated in the `RefreshProtocol` type.

ckks/keygen.go
```diff
@@ -101,21 +102,39 @@ func (keygen *keyGenerator) GenSecretKey() (sk *SecretKey) {
 
 func (keygen *keyGenerator) GenSecretKeyGaussian(sigma float64) (sk *SecretKey) {
 	sk = new(SecretKey)
-	sk.sk = keygen.ckksContext.contextQP.SampleGaussianNTTNew(sigma, uint64(6*sigma))
+	prng, err := utils.NewPRNG()
+	if err != nil {
+		panic(err)
+	}
+	gaussianSampler := ring.NewGaussianSampler(prng, keygen.ckksContext.contextQP)
```
ChristianMct, line 109: Allocation of sampler in type.

ckks/keygen.go
```diff
@@ -101,21 +102,39 @@ func (keygen *keyGenerator) GenSecretKey() (sk *SecretKey) {
...
 
 // GenSecretKeyWithDistrib generates a new SecretKey with the distribution [(p-1)/2, p, (p-1)/2].
 func (keygen *keyGenerator) GenSecretKeyWithDistrib(p float64) (sk *SecretKey) {
 	sk = new(SecretKey)
-	sk.sk = keygen.ckksContext.contextQP.SampleTernaryMontgomeryNTTNew(p)
+	prng, err := utils.NewPRNG()
+	if err != nil {
+		panic(err)
+	}
+	ternarySampler := ring.NewTernarySampler(prng, keygen.ckksContext.contextQP)
```
ChristianMct, line 122: Allocation of sampler in type.

bfv/encryptor.go
```diff
@@ -170,29 +171,38 @@ func (encryptor *pkEncryptor) encrypt(plaintext *Plaintext, ciphertext *Cipherte
 
 	var ringContext *ring.Context
 
+	prng, err := utils.NewPRNG()
```
ChristianMct, line 174: Allocation of sampler should be at the creation of the `Encryptor` struct.

ring/ternarySampler.go
```diff
@@ -180,59 +190,55 @@ func (context *Context) sampleTernary(samplerMatrix [][]uint64, p float64, pol *
 
 		matrix := computeMatrixTernary(p)
```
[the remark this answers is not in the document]
Pro7ech: The pre-computation of the matrix is negligible (only 3 row). We might need for various p, Juan openly asked me for it. I'd leave it like that.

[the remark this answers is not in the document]
ChristianMct: Well, more than the computation cost, I see some calls to `make` which can be avoided. 

Regarding the values for `p`, we have to characterize the exact use of what Juan mentioned: it seems unlikely for me that the *same* sampler instance needs several values for `p` (e.g., an `Encryptor` type needs a fixed `p` for all its output). In that case, creating a new `TernarySampler` for each new `p` is reasonable.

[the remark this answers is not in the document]
ChristianMct: @eliedz We had a discussion with @Pro7ech and decided to go for a reduction of the samplers' interface by allowing the user to initialize a sampler according to its needs and then simply call a `Read` method on it (as previously discussed). 

I had a first shot at it for the `TernarySampler` and it looks good. So I suggest we proceed with the `GaussianSampler` the same way. Let me know what you think.

ring/sampler_gaussian.go
```diff
@@ -19,10 +19,49 @@ func NewGaussianSampler(prng utils.PRNG, context *Context) *GaussianSampler {
...
 	gaussianSampler.prng = prng
 	gaussianSampler.randomBufferN = make([]byte, context.N)
+	gaussianSampler.sample = gaussianSampler.SampleLvl
 	return gaussianSampler
 }
 
-// Sample samples a truncated gaussian polynomial with variance
+// NewGaussianSampler creates a new instance of GaussianSampler.
+// Accepts a PRNG and context and samples different kinds of gaussian polynomials
+func NewGaussianSamplerLvl(prng utils.PRNG, context *Context) *GaussianSampler {
```
ChristianMct, line 28: To follow the same pattern as the other samplers, the `NewGaussianSampler` constructors should accept the `sigma` and `bound` parameters and they should be removed from the interface. 

@Pro7ech Correct me if I'm wrong, I think we agreed that the loss in flexibility was acceptable wrt streamlining the interface.

ring/sampler_uniform.go
```diff
@@ -67,22 +66,22 @@ func (uniformSampler *UniformSampler) Sample(Pol *Poly) {
...
-	uniformSampler.Sample(Pol)
+	uniformSampler.Read(Pol)
 
 	return
 }
 
-// SampleLvlNew generates a new polynomial with coefficients following a uniform distribution over [0, Qi-1].
-func (uniformSampler *UniformSampler) SampleLvlNew(level uint64) (Pol *Poly) {
+// ReadNewLvl generates a new polynomial with coefficients following a uniform distribution over [0, Qi-1].
+func (uniformSampler *UniformSampler) ReadNewLvl(level uint64) (Pol *Poly) {
```
ChristianMct, line 80: This method crashes (I pushed a test case for it).

ChristianMct: After discussing with @Pro7ech, we can completely remove this method.

PR #51, opened 2020-06-17
dckks: add missing err

tharvik:
when running `go build ./...`, `dckks` fails to build, making this package impossible to import. Adding a declaration for `err`, it works like a charm.


ChristianMct:
This bug was fixed in by #47 but merging this one nevertheless.

PR #52, opened 2020-06-17
Improve travis & Makefile

tharvik:
Depends on #51 

* travis should build with the latest go version, it now do so, with 1.13 & 1.14 being the two supported version
* travis now can catch issues similar to #51
* the Makefile looked like an old copy of the one in dedis/Coding, using now upstream direclty
* fix the lint issues found by the new version of dedis/Coding

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ _testmain.go
 # vendor/
 
 .idea/
+
+# generated
+/Coding
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,10 @@
 language: go
 
 go:
+  - 1.x
+  - 1.14.x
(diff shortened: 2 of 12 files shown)


ChristianMct:
Is the envisioned roadmap for github.com/dedis/Coding aware of external software ? Are changes frequently expected ?

tharvik:
> Is the envisioned roadmap for github.com/dedis/Coding aware of external software ? Are changes frequently expected ?

It is used in multiple project at EPFL, as such, changes are rare to avoid any CI failure. Usually, when I'm improving it, if anything break, I'm making PRs to the project in needs.

ChristianMct:
Thanks for the clarification.

PR #60, opened 2020-09-07
Clean the Ring interface

ChristianMct:
The goal is to improve the interface of the `lattigo/ring` package. The main concern is to avoid the generic term of "Context" to designate the polynomial ring.


Pro7ech (approved):

Pro7ech:
Included in v2.0.0

PR #67, opened 2020-10-20
Lazy reduction and low level optimizations

Pro7ech:
Added lazy reduction to NTT and basis-extension. Optimized several low level arithmetic in Ring, notably basis extension and division in the RNS domain. Overall gain is ~15% on the CKKS (and BFV) keyswitch.

Changed the default moduli in Ring to 61-bits moduli (from 60) to have test using the maximum size moduli.

With butterflies (x,y)->(x+y * psi, x-y*psi), the NTT now applies a conditional reduction on the input x>=4q every odd level. We need to consider two cases : the absolute worst case growth 1) (y * psi)%q = 2q-1 per level, and the worst case growth that does not trigger the reduction 2) x+(y * psi)%q = 4q-2. 

If the input of the NTT is bounded by q-1: 1) q-1 -> 3q-2 -> (5q-3 -> q-3) -> 3q-3 and 2) q-1 -> 3q-2 -> (4q-2 -> 4q-2) -> 6q-3 -> (8q-4 -> 4q-4) -> 6q-4.
If the input to the NTT is bounded by 2q-1 : 1) 2q-1 -> 4q-2 -> (6q-3 -> 4q-3) -> 6q-4  and 2) 2q-1 -> 4q-2 -> (4q-2 -> 4q-2) -> 6q-3 > (8q-4 -> 4q-4) -> 6q-4.

Hence all worst cases the values never go above 8q, and since we use 61 bits moduli (at most) we can afford up to 8q growth.


ckks/bootstrapp_test.go
```diff
@@ -12,12 +13,20 @@ import (
...
+		t.Skip("skipping bootstrapping tests for GOARCH=wasm")
+	}
+
 	rand.Seed(time.Now().UnixNano())
 
 	var err error
 	var testContext = new(testParams)
 
-	paramSet := uint64(0)
+	paramSet := uint64(1)
```
ChristianMct, line 29: We don't want to test the smallest set ?

Makefile
```diff
@@ -18,7 +18,8 @@ test_examples:
 
 .PHONY: test_gotest
 test_gotest:
-	go test -v -short -p=1 ./... -timeout=0
+	go test -v -short -p=1 -timeout=0 ./utils ./ring ./bfv ./dbfv ./dckks
+	go test -v -short -p=1 -timeout=0 ./ckks -test-bootstrapping
```
ChristianMct, line 22: I changed this flag name to "test-bootstrapping" so that it is more explicit

ckks/bootstrapp.go
```diff
@@ -14,12 +13,27 @@ func (btp *Bootstrapper) Bootstrapp(ct *Ciphertext) *Ciphertext {
...
 	var ct0, ct1 *Ciphertext
 
-	for ct.Level() != 0 {
+	for ct.Level() != 1 {
 		btp.evaluator.DropLevel(ct, 1)
 	}
 
 	// Brings the ciphertext scale to Q0/2^{10}
-	btp.evaluator.ScaleUp(ct, math.Round(btp.prescale/ct.Scale()), ct)
+	if ct.Level() == 1 {
```
ChristianMct, line 21: This behavior needs to be documented in the GoDoc above.

ckks/bootstrapp_test.go
Pro7ech: The -short flag will tune down N and n, regardless of the set, so it doesn't matter what is tested.

ckks/bootstrapp.go
Pro7ech: I updated the godoc

ChristianMct (approved):

bfv/bfv_benchmark_test.go
```diff
@@ -74,19 +74,19 @@ func benchEncrypt(testctx *testContext, b *testing.B) {
 	plaintext := NewPlaintext(testctx.params)
 	ciphertext := NewCiphertextRandom(testctx.prng, testctx.params, 1)
 
-	b.Run(testString("Encrypt/Pk/Slow", testctx.params), func(b *testing.B) {
+	b.Run(testString("Encrypt/Pk/Slow/", testctx.params), func(b *testing.B) {
```
ChristianMct, line 77: If it was just "slow", we would not be using it. Since this refers to the "usual" BFV encryption, it could just be `Encrypt/Pk`.

Also, if we follow the notation that we had for parameters until now, it could be `Encrypt/key=pk` (would need to be reflected on the CKKS side). What do you think ?

ChristianMct (changes requested):

ChristianMct (approved):

PR #69, opened 2020-10-26
REAL-HEAAN

Pro7ech:
Implementation of the work of Duhyeong Kim and Yongsoo Song (https://link.springer.com/chapter/10.1007/978-3-030-12146-4_6).

Unlike the original CKKS, which encrypts vectors of N/2 complex numbers, their variant of the scheme is able to encrypt vectors of N real numbers.

The scheme is based on the properties of the subring Z[X + X^{1}]/(X^{2N} + 1) of Z[X]/(X^{2N} + 1), named the conjugate invariant ring, for which the R-LWE problem can be shown to be as hard as the in ring Z[X]/(X^{N} + 1). 

Using a linear map from Z[X + X^{1}]/(X^{2N} + 1) to Z[X]/(X^{N} + 1) we can operate in this sub-ring with a complexity nearly identical to the one of the ring Z[X]/(X^{N} + 1).

This implementation slightly deviates from their work to be able to re-use the already implemented algorithms of Lattigo. Instead of using a mapping Z[X + X^{1}]/(X^{2N} + 1) ->  Z[X]/(X^{N} -1) (cyclic convolution), it uses a mapping Z[X + X^{1}]/(X^{2N} + 1) -> Z[X]/(X^{N} + 1) (nega-cyclic convolution).


crockeea:
What's the status on this PR? I would love to see this in Lattigo!

ChristianMct:
This branch is functional but will not be merged as such because it re-implements too much of what is common to all schemes in Lattigo. Hence, it is on hold until we complete a series of feature extraction that implement this common basis.

PR #71, opened 2020-10-30
EvaluatePoly/Cheby to return an error when ciphertext level doesn't enable full evaluation

Pro7ech:
evaluatePoly/Cheby will now return an error if the ciphertext does not have enough level to evaluate the polynomial. Added more errors that can be returned when trying to rescale with invalid scalings.


ckks/polynomial_evaluation.go
```diff
@@ -25,6 +26,43 @@ func NewPoly(coeffs []complex128) (p *Poly) {
...
+		valueInt := int64(cImag)
+		valueFloat := cImag - float64(valueInt)
+
+		if valueFloat != 0 {
+			needScale = true
+		}
+	}
+
+	logDegree := uint64(math.Log2(float64(len(pol.coeffs))) + 0.5)
+	if needScale {
```
ChristianMct, line 55: All the function until this line seems to be replaceable by:
`if real(c) != float64(int64(real(c))) || imag(c) != float64(int64(imag(c))) {
logDegree ++
}`

ChristianMct (changes requested):

ckks/evaluator.go
```diff
@@ -1247,6 +1247,18 @@ func (eval *evaluator) Rescale(ct0 *Ciphertext, threshold float64, ctOut *Cipher
 
 	ringQ := eval.ringQ
 
+	if threshold == 0 {
+		return errors.New("cannot Rescale: threshold is 0")
```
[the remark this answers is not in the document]
Pro7ech: Fixed

ckks/polynomial_evaluation.go
Pro7ech: fixed

ChristianMct (approved):
To be merged with squash&merge.

PR #74, opened 2020-11-04
[bfv] added special-purpose plaintexts encodings

Pro7ech:
Changed NewPlaintext to : 

- NewPlaintext : calls NewPlaintextZQ.

- NewPlaintextZQ : encodes the plaintext in ZQ and scales it by Q/t. Slower encoding and larger plaintext size, but faster encryption and faster ct + pt additions. Very slow ct * pt.

- NewPlaintextZT : encodes the plaintext in ZT (without scaling by Q/t). Faster encoding and smaller plaintext size, but slower encryption and slower ct + pt additions. Slower ct * pt than NewPlaintextMul but faster than NewPlaintextZQ.

- NewPlaintextMul : encode the plaintext in the NTT and Montgomery domain of ZQ (without scaling by Q/t). Very fast ct * pt, cannot do ct + pt.

QiMul does not need to be given anymore in the params, and is automatically generated by the evaluator.

ring : GenNTTPrimes now takes the Nth primitive root as input rather than logN.


CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 # Changelog
 All notable changes to this project will be documented in this file. 
 
+
+## [2.0.x] - 2020-12-02
+
+### Changed
+- BFV : plaintext must not be created using NewPlaintextZQ, NewPlaintextZT or NewPlaintextMul. See bfv/plaintext.go for the full details.
+- BFV : QiMul is now automatically generated by the evaluator
```
ChristianMct, line 10: The important thing is not where it is generated , but that it is controlled by the user in a different way.

I would simply add that it is computed from X and Y after the corresponding entry in the "Removed" section below.
Also please  specify use the proper type names to improved readability. E.g., removed field `QiMul` from `bfv.Parameters`, it is now computed directly from X and Y

bfv/encryptor.go
```diff
@@ -293,5 +304,38 @@ func (encryptor *skEncryptor) encrypt(plaintext *Plaintext, ciphertext *Cipherte
...
+	} else {
+		encryptor.ringQ.Add(ciphertext.value[0], plaintext.value, ciphertext.value[0])
+	}
+}
+
+func (encryptor *encryptor) TtoQ(p *Plaintext) {
+
+	ringQ := encryptor.ringQ
+	encryptor.polypool[0].Zero()
+	for i := len(ringQ.Modulus) - 1; i >= 0; i-- {
```
ChristianMct, line 319: Shouldn't this operation be part of the `ring` package ?

bfv/encryptor.go
```diff
@@ -293,5 +304,38 @@ func (encryptor *skEncryptor) encrypt(plaintext *Plaintext, ciphertext *Cipherte
...
-	ringQ.Add(ciphertext.value[0], plaintext.value, ciphertext.value[0])
+	if !plaintext.inZQ {
+		encryptor.TtoQ(plaintext)
+		encryptor.ringQ.Add(ciphertext.value[0], encryptor.polypool[0], ciphertext.value[0])
+	} else {
+		encryptor.ringQ.Add(ciphertext.value[0], plaintext.value, ciphertext.value[0])
+	}
+}
+
+func (encryptor *encryptor) TtoQ(p *Plaintext) {
```
ChristianMct, line 315: Should this method check that `!p.inZQ` ?

bfv/evaluator.go
```diff
@@ -160,14 +167,63 @@ func (eval *evaluator) getElemAndCheckUnary(op0, opOut Operand, opOutMinDegree u
 	return // TODO: more checks on elements
 }
 
+func (eval *evaluator) TtoQ(p *Plaintext, pol *ring.Poly) {
```
ChristianMct, line 170: Seems like a duplicated implementation here. See my other comment.

ChristianMct (commented):
We need to further polish the plaintext encoding API and semantic.

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 # Changelog
 All notable changes to this project will be documented in this file. 
 
+
+## [2.0.x] - 2020-12-02
+
+### Changed
+- BFV : plaintext must not be created using NewPlaintextZQ, NewPlaintextZT or NewPlaintextMul. See bfv/plaintext.go for the full details.
```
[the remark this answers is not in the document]
Pro7ech: Yes its a typo, I fixed it

bfv/encryptor.go
Pro7ech: To be seen at a later stage if we want to further generalize the operations to vectors, but this would require some work directly on the ring package. For now I'd leave like this.

bfv/encryptor.go
Pro7ech: I made those method private, as the check is done before.

bfv/evaluator.go
Pro7ech: made tToQ private and generale

ChristianMct (approved):

PR #80, opened 2020-11-12
[ckks] API improvements and documentation

Pro7ech:
API changes :

- evaluator.RotateColumns -> Evaluator.Rotate
- evaluator.EvaluateChebySpecial -> removed
- evaluator.EvaluateCheby -> the change of variable isn't done automatically anymore and the user must do it before calling the function to ensure correctness. The reason is that additional constant might be merged with the change of variable and/or the change of variable might be merged with previous operations. This leaves more freedom to the user to optimize a circuit. The method also doesn't take anymore into account the change of variable for checking if the number of remaining levels is sufficient to evaluate the polynomial.
- encoder : when encoding, the number of slots must now be given in log2 basis. This is to prevent errors that would induced by zero values or non power of two values.
- encoder : added EncodeAtLvlNew and EncodeNTTAtLvlNew API

- Updated change-log to reflect the aforementioned changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 # Changelog
 All notable changes to this project will be documented in this file. 
 
+## [2.0.x] - 2020-11-20
+
+### Changed
+- CKKS : API of evaluator.RotateColumns becomes Evaluator.Rotate
+- CKKS : the change of variable in evaluator.EvaluateCheby isn't done automatically anymore and the user must do it before calling the function to ensure correctness. The reason is that additional constant might be merged with the change of variable and/or the change of variable might be merged with previous operations. This leaves more freedom to the user to optimize a circuit. The method also doesn't take anymore into account the change of variable for checking if the number of remaining levels is sufficient to evaluate the polynomial.
+- CKKS : when encoding, the number of slots must now be given in log2 basis. This is to prevent errors that would induced by zero values or non power of two values.
+- CKKS : new encoder API : EncodeAtLvlNew and EncodeNTTAtLvlNew, which allow a user to encode a plaintext at a specific level.
+
+### Removed
+- CKKS : evaluator.EvaluateChebySpecial
+
 ## [2.0.0] - 2020-10-07
 
 ### Performance
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 - CKKS : API of evaluator.RotateColumns becomes Evaluator.Rotate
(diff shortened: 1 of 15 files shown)


ckks/polynomial_evaluation.go
```diff
@@ -110,44 +110,14 @@ func (eval *evaluator) EvaluatePoly(ct0 *Ciphertext, pol *Poly, evakey *Evaluati
 // Returns an error if something is wrong with the scale.
 func (eval *evaluator) EvaluateCheby(op *Ciphertext, cheby *ChebyshevInterpolation, evakey *EvaluationKey) (opOut *Ciphertext, err error) {
 
-	if err := checkEnoughLevels(op.Level(), &cheby.Poly, 2/(cheby.b-cheby.a)); err != nil {
+	if err := checkEnoughLevels(op.Level(), &cheby.Poly, 1); err != nil {
```
ChristianMct, line 113: Please explain this line in a comment. Why did it went from `2/(cheby.b-cheby.a)` to `1` ?

Par of the challenge of understanding what is going on here is that `checkEnoughLevels` itself is undocumented. Can you take care of that ? Side note: shouldn't `checkEnoughLevels`take an `Operand` as input ?

ChristianMct (changes requested):

ChristianMct:
Also, can you create a new "unreleased" section at the top of the Changelog and summarize the important changes (especially) there ? In particular, those changes that change the public API.

ckks/polynomial_evaluation.go
Pro7ech: Previously it was checking if the change of variable would consume a level (i.e. if 2/(b-a) wasn't an integer). Since the user now has to do the change of variable himself, this check doesn't have to be done. Rather than entirely removing the check, I pass the value 1 (an integer), like it is done for the evaluation with a standard basis.

CHANGELOG.md
```diff
@@ -2,6 +2,17 @@
 # Changelog
 All notable changes to this project will be documented in this file. 
 
+## [2.0.x] - 2020-11-20
+
+### Changed
+- CKKS : API of evaluator.RotateColumns becomes Evaluator.Rotate
+- CKKS : the change of variable in evaluator.EvaluateCheby isn't done automatically anymore and the user must do it before calling the function to ensure correctness. The reason is that additional constant might be merged with the change of variable and/or the change of variable might be merged with previous operations. This leaves more freedom to the user to optimize a circuit. The method also doesn't take anymore into account the change of variable for checking if the number of remaining levels is sufficient to evaluate the polynomial.
```
ChristianMct, line 9: Please keep entries in the changelog short. What you describe here belongs to the godoc.

ckks/polynomial_evaluation.go
ChristianMct: I cannot locate this comment on the last commit. What about the second part of my comment about `checkEnoughLevels` and its arguments ?

Pro7ech: I documented the checkEnoughLevels method

CHANGELOG.md
Pro7ech: reduced the change size

ChristianMct (approved):

PR #90, opened 2020-12-22
Dev v2.1.0 code review

Pro7ech:
cf. changelog


bfv/operand.go
```diff
@@ -82,23 +82,6 @@ func (el *Element) Degree() uint64 {
 	return uint64(len(el.value) - 1)
 }
 
-// Resize resizes the target Element degree to the degree given as input. If the input degree is bigger, then
-// it will append new empty polynomials; if the degree is smaller, it will delete polynomials until the degree matches
-// the input degree.
-func (el *Element) Resize(params *Parameters, degree uint64) {
```
ChristianMct, line 88: This was a public function. Is there a specific reason for removing it beside that its not used/tested ?

bfv/params.go
```diff
@@ -13,24 +13,15 @@ import (
 // MaxLogN is the log2 of the largest supported polynomial modulus degree.
 const MaxLogN = 16
 
+// MinLogN is the log2 of the smallest supported polynomial modulus degree.
+const MinLogN = 4
```
ChristianMct, line 17: Why not 1 ? Also it should be explicit that this is the minimum supported log2 for correctness of the operations, but that such small instances are insecure.

ckks/params.go
```diff
@@ -14,6 +14,9 @@ import (
 // MaxLogN is the log2 of the largest supported polynomial modulus degree.
 const MaxLogN = 16
 
+// MinLogN is the log2 of the smallest supported polynomial modulus degree
+const MinLogN = 4
```
ChristianMct, line 18: Same as for BFV comment

examples/dbfv/psi/main.go
```diff
@@ -106,17 +105,17 @@ func main() {
 	ternarySamplerMontgomery := ring.NewTernarySampler(prng, ringQP, 0.5, true)
 
 	// Create each party, and allocate the memory for all the shares that the protocols will need
-	P := make([]*party, N, N)
```
ChristianMct, line 109: For consistency, shouldn't the PSI example follow the same structure as the PIR one ?

ChristianMct:
We need the appropriate CHANGELOG.md entry(-ies) to be added.

bfv/params.go
Pro7ech: Minimum must be 2^4 to ensure the NTT correctness.

ckks/params.go
Pro7ech: Minimum must be 2^4 to ensure the NTT correctness.

examples/dbfv/psi/main.go
Pro7ech: Yes we need to reduce its cyclomatic complexity too

bfv/operand.go
Pro7ech: Following the code review suggestion, unused code should be removed. As far as I am concerned, this function has never been used and has no real purpose anymore since the evaluator now enforces the output ciphertext to be of the correct size.

ChristianMct: It might be used in user code, since `Resize` is an exported function of e.g., `Ciphertext`.

If we remove it, we need to have a line of the changelog about that.

bfv/params.go
ChristianMct: Ok for me. Mention this in the comment too.

ckks/params.go
ChristianMct: Ok for me. Mention this in the comment too.

bfv/operand.go
Pro7ech: We will keep it, to make it symmetric with CKKS. In addition, the .Level() method has been added and might be needed in the future.

bfv/params.go
Pro7ech: Done

ckks/params.go
Pro7ech: Done

examples/dbfv/psi/main.go
Pro7ech: Done

ChristianMct (changes requested):
The README still needs some polishing. Otherwise all good for me.

ChristianMct (approved):

PR #93, opened 2021-01-08
Dev vector ops

Pro7ech:
[ckks] : added efficient vector manipulation API (linear transforms) that can be used as building blocks to do matrix x matrix operations.
[ckks] : removed evaluator.RescaleMany. It is now merged with evaluator.Rescale which supports optimized many rescaling.
[ckks] : EvaluatePoly and EvaluateCheby now take the target scale as input.
[ckks] : complete redesign of the bootstrapping parameters, which allows for an easier and much fine grained parameterizatiom.
[ckks] : the bootstrapping can now be parameterized with the ratio between the modulus and the message, as well as composing arcsine to improve the precision when the ratio is small.

[ring] : changed poly.GetDegree() -> poly.N()
[ring] : changed poly.GetLenModuli() -> poly.Level() (replace by poly.Level()+1 to get the old value)
[ring] : all API starting by DivByLastModulus is now  in-place and takes an output target.


ChristianMct:
Changes have been integrated in another branch.

PR #125, opened 2021-06-07
Issue#124

Pro7ech:
Also changed the internal pools of the encoder to be reset before being used, not after.

### DIFF
--- a/ckks/encoder.go
+++ b/ckks/encoder.go
@@ -166,7 +166,6 @@ func (encoder *encoderComplex128) EncodeAtLvlNew(level int, values []complex128,
 func (encoder *encoderComplex128) Encode(plaintext *Plaintext, values []complex128, logSlots int) {
 	encoder.Embed(values, logSlots)
 	encoder.ScaleUp(plaintext.value, plaintext.scale, encoder.ringQ.Modulus[:plaintext.Level()+1])
-	encoder.WipeInternalMemory()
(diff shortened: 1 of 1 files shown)


ckks/encoder.go
```diff
@@ -232,11 +233,18 @@ func (encoder *encoderComplex128) GetErrSTDSlotDomain(valuesWant, valuesHave []c
 // GetErrSTDCoeffDomain returns the scaled standard deviation in the coefficient domain of the difference between two complex vectors in the slot domains
 func (encoder *encoderComplex128) GetErrSTDCoeffDomain(valuesWant, valuesHave []complex128, scale float64) (std float64) {
 
+	encoder.WipeInternalMemory()
+
+	nextPow2 := 1
+	for nextPow2 < len(valuesWant) {
+		nextPow2 <<= 1
+	}
```
ChristianMct, line 241: Can this loop be replaced by `1 << bits.Len64(uint64(x-1))` ? 

I agree that a loop-free version is less explicit but you can  compensate with a code comment.

Pro7ech: It can, but then the file would need to import math/bits just for that line.

ChristianMct: I don't see any problem in importing `math/bits`. What would it be ?

Pro7ech: Made the required changes

ChristianMct (approved):

PR #128, opened 2021-06-09
added vOLE example & fixed bug in ring/uniform sampler

Pro7ech:
Todo : 

- [x] Document
- [x] Update makefile


examples/ring/vOLE/main.go
```diff
@@ -0,0 +1,406 @@
+package main
```
ChristianMct, line 1: There needs to be an extensive godoc that describes this example and provides links to the paper for further reference.

examples/ring/vOLE/main.go
```diff
@@ -0,0 +1,406 @@
...
+	for i := range lns.coeffs {
+		lns.coeffs[i] = ring.RandInt(bound)
+	}
+
+	lns.baseRing.SetCoefficientsBigint(lns.coeffs, pol)
+
+	return
+}
+
+func main() {
```
ChristianMct, line 100: The example needs to be added to the `Makefile`'s `test_example` target.

ChristianMct (changes requested):

examples/ring/vOLE/main.go
Pro7ech: Right!

jrtroncoso (approved):
I've made a few changes in the vOLE example file. The rest of the updates look good

ChristianMct (approved):

PR #130, opened 2021-06-10
Dev mkhe

hedi-sassi:
Addition of a multi-key encryption scheme based on BFV and CKKS.


mkckks/evaluator.go
```diff
@@ -0,0 +1,472 @@
...
+// evaluateInPlace performs an evaluation of the evaluate function between c0 and c1, and stores the result into ctOut by optimizing the computation if there are nils. Subtraction needs to be handled ad-hoc and is indicated with isSub bool
+func (eval *mkEvaluator) evaluateInPlace(c0, c1, ctOut *ckks.Ciphertext, isSub bool, evaluate func(uint64, *ring.Poly, *ring.Poly, *ring.Poly)) {
+
+	var tmp0, tmp1 *ckks.Element
+	level := utils.MinUint64(utils.MinUint64(c0.Level(), c1.Level()), ctOut.Level())
+
+	minDegree := utils.MinUint64(c0.Degree(), c1.Degree())
+
+	if c1.Scale() > c0.Scale() {
+		tmp0 = new(ckks.Element)
```
andru47, line 431: I have been using the multi-key functionality implemented here for CKKS for the last two days and I think there is a problem with this allocation. In my example, I had two ciphertexts with different scales that were added together and this would throw errors. The error is thrown when you try to access the Level() on the passed object created here and it's thrown because the Value array from the element is empty and the code tries to access the element in position 0. My solution was to call `MultByConstNew` in the next if statement and copy the value to tmp0, but maybe the tmp0 initialisation can be changed.

hedi-sassi: Hello. You're completely right. The same problem should arise for the tmp1 variable at line 440. In the `ckks.evaluator`, their implementation of `evaluateInPlace` uses a ciphertext pool with pre initialized ciphertexts so it can be fixed that way. Else, calling `MultByConstNew` will effectively correct the bug.

Maybe you could ask Christian or Jean-Phillipe what would be the best option.

ChristianMct: Hello @andru47.
Thanks for reporting this finding. We'll fix this bug soon as we will soon try to merge this branch with the master branch.

mkrlwe/marshaler.go
```diff
@@ -0,0 +1,332 @@
...
+		}
+	}
+	swk.Value = value
+	evalKey.Key01 = swk
+
+	// Unmarshal the decomposedPoly
+	k2 := new(MKDecomposedPoly)
+	dpSize := binary.LittleEndian.Uint64(data[dpStartIndex : dpStartIndex+8])
+	dpoly := make([]*ring.Poly, dpSize)
+	pointer = dpStartIndex
```
andru47, line 146: Hi. Marshalling and un-marshalling a mk relinearization key produces errors for the field `Key2`. I think the pointer here should take value `dpStartIndex + 8`, so it can omit the bytes that indicate the length of the encoded string.

mkrlwe/marshaler.go
```diff
@@ -0,0 +1,332 @@
...
+	evalKey.PeerID = pid
+
+	return nil
+
+}
+
+// MarshalSwitchingKey encodes a switching in a byte slice.
+func MarshalSwitchingKey(swk *rlwe.SwitchingKey) (data []byte, err error) {
+	dataLen, numEl := GetSwkDataLen(swk, true)
+	data = make([]byte, dataLen+7)
```
andru47, line 172: Not really sure if I am correct, but I think it should be `+ 8`, since `binary.LittleEndian.PutUint64` takes 8 bytes. Same on line 262.

ChristianMct:
Since MKHE is not part of the current backlog, we're closing this PR. The `dev_mkhe` branch will not be deleted but cannot be merged in the current state (it has diverged too much from main branch and from the current state-of-the-art in MKHE).

PR #131, opened 2021-06-18
Centralizing RLWE key related code across schemes

Pro7ech:
### CKKS/BFV
- Moved and centralized key-generation to `rlwe/keygen.go`
- Moved and centralized key-switching to `rlwe/keyswitch.go`
- Moved and centralized encryption to `rlwe/encryptor.go`
- Moved and centralized decryption to `rlwe/decryptor.go`

### BFV
- `bfv.Paramters` now holds a single read-only copies of `ringT` and `ringQMul`. Similarely to `rlwe.Parameters`, these rings are never copied, but given as pointer to other bfv structs.
- The moduli of `ringQMul` are now allocated based on _N_ and _Q_.

### RLWE
- `rlwe.Parameters` now generates and stores its own `ring.Ring` instances, **which are read-only**. They are now meant to be instantiated once per scheme instance and passed as pointers to the different structs. This change further reduces the memory footprint of all scheme related `structs`, for which the memory pool should now be the dominant use of memory
- Removed `IsNTT` flag from `rlwe.Elements` and `rlwe.Plaintext`. They are now part of `ring.Poly`
- Added generalized code for RLWE key generation, encryption and decryption with respective tests

### Ring
- `ring.Ring` is now read-only (no method called through the API of the `ring` package will modify in an instantiated `ring.Ring`, but the public field can still be accessed and modified manually)
- Added `IsNTT` and `IsMForm` public flags in `ring.Poly`. These flags **are never checked or modified** by methods in `ring` package  and are **only transferred to other polynomials in exceptional cases and if explicitly stated**. They are meant to be used to facilitate the tracking of the domain of a `ring.Poly` when building schemes on top of `Ring`
- Added `CopyValues` API family to copy values of a polynomial without copying `IsNTT` and `IsMForm` flags
- Updated Godoc to reflect if `IsNTT` and `IsMForm` are transferred when a copy of a polynomial is made
- All methods that copy values are now faster
- Further code footprint reduction
- Removed `MulPoly` family API and tests (this API was not used and was a simple composition of already existing API)


bfv/encryptor.go
```diff
@@ -3,279 +3,51 @@ package bfv
...
 	"github.com/ldsec/lattigo/v2/ring"
 	"github.com/ldsec/lattigo/v2/rlwe"
-	"github.com/ldsec/lattigo/v2/utils"
 )
 
-// Encryptor in an interface for encryptors
-//
-// encrypt with pk : ciphertext = [pk[0]*u + m + e_0, pk[1]*u + e_1]
-// encrypt with sk : ciphertext = [-a*sk + m + e, a]
-type Encryptor interface {
```
ChristianMct, line 13: It is probably better to keep an interface at the scheme's level too.

ckks/encryptor.go
```diff
@@ -3,394 +3,54 @@ package ckks
...
 	"github.com/ldsec/lattigo/v2/ring"
 	"github.com/ldsec/lattigo/v2/rlwe"
-	"github.com/ldsec/lattigo/v2/utils"
 )
 
-// Encryptor in an interface for encryptors
-//
-// encrypt with pk : ciphertext = [pk[0]*u + m + e_0, pk[1]*u + e_1]
-// encrypt with sk : ciphertext = [-a*sk + m + e, a]
-type Encryptor interface {
```
ChristianMct, line 13: It is probably better to keep an interface at the scheme's level.

bfv/decryptor.go
```diff
@@ -1,69 +1,25 @@
 package bfv
 
 import (
-	"github.com/ldsec/lattigo/v2/ring"
 	"github.com/ldsec/lattigo/v2/rlwe"
 )
 
-// Decryptor is an interface for decryptors
-type Decryptor interface {
```
ChristianMct, line 9: It is probably better to keep an interface at the scheme's level too.

bfv/keygen.go
```diff
@@ -1,254 +1,36 @@
...
-	"math/big"
-
-	"github.com/ldsec/lattigo/v2/ring"
 	"github.com/ldsec/lattigo/v2/rlwe"
-	"github.com/ldsec/lattigo/v2/utils"
 )
 
 // KeyGenerator is an interface implementing the methods of the keyGenerator.
 type KeyGenerator interface {
 	GenSecretKey() (sk *rlwe.SecretKey)
```
ChristianMct, line 9: Can probably be rewritten as something like:

```
type KeyGenerator interface {
	rlwe.KeyGenerator
        <specific methods>
}
```

ckks/decryptor.go
```diff
@@ -1,79 +1,26 @@
...
 
 import (
-	"github.com/ldsec/lattigo/v2/ring"
 	"github.com/ldsec/lattigo/v2/rlwe"
-	"github.com/ldsec/lattigo/v2/utils"
 )
 
-// Decryptor is an interface for decrypting Ciphertexts. A Decryptor stores the secret-key.
-type Decryptor interface {
-	// DecryptNew decrypts the ciphertext and returns a newly created
```
ChristianMct, line 11: It is probably better to keep an interface at the scheme's level too.

ckks/encryptor.go
```diff
@@ -3,394 +3,54 @@ package ckks
...
-		ringQ.NTTLvl(lvl, ciphertext.Value[0], ciphertext.Value[0])
-		ringQ.NTTLvl(lvl, ciphertext.Value[1], ciphertext.Value[1])
-
-		if plaintext.IsNTT {
-			// ct0 = (u*pk0 + e0)/P + m
-			ringQ.AddLvl(lvl, ciphertext.Value[0], plaintext.Value, ciphertext.Value[0])
-		}
+// NewEncryptor instanciate a new wrapped rlwe.Encryptor for CKKS ciphertexts and plaintexts.
+func NewEncryptor(params Parameters, key interface{}) *Encryptor {
+	switch key := key.(type) {
```
ChristianMct, line 15: Can this switch be done at the rlwe layer ?

ckks/keygen.go
```diff
@@ -28,248 +24,16 @@ type KeyGenerator interface {
 // KeyGenerator is a structure that stores the elements required to create new keys,
 // as well as a small memory pool for intermediate values.
 type keyGenerator struct {
```
ChristianMct, line 26: Let's discuss if needed. We could have a `GenRelinearizationKeyForCiphertextDegree` method in the `rlwe.KeyGenerator` interface and it would re-align the two schemes.

rlwe/decryptor.go
```diff
@@ -0,0 +1,109 @@
...
+	// DecryptNew decrypts the ciphertext and returns a newly created
+	// plaintext. A Horner method is used for evaluating the decryption.
+	// The level of the output plaintext is ciphertext.Level().
+	DecryptNew(ciphertext *Element) (plaintext *Plaintext)
+
+	// DecryptNew decrypts the ciphertext and returns a newly created
+	// plaintext. A Horner method is used for evaluating the decryption.
+	// The level of the output plaintext is ciphertext.Level().
+	// Plaintext output is in the NTT domain.
+	DecryptNTTNew(ciphertext *Element) (plaintext *Plaintext)
```
ChristianMct, line 19: This is only for CKKS right ? Wouldn't it be preferable to simply have a final NTT in the `ckks.Decryptor.Decrypt` `?

rlwe/encryptor.go
```diff
@@ -0,0 +1,509 @@
...
+	poolP1 := encryptor.poolP[1]
+	poolP2 := encryptor.poolP[2]
+
+	// We sample a R-WLE instance (encryption of zero) over the extended ring (ciphertext ring + special prime)
+
+	ringQ := encryptor.ringQ
+
+	ciphertextNTT := ciphertext.Value[0].IsNTT
+
+	if fast {
```
ChristianMct, line 257: I don't know why we changed from having two different functions. It would look much better to have both a `encrypt` and `encryptFast` methods.

rlwe/encryptor.go
```diff
@@ -0,0 +1,509 @@
...
+	// encrypting zero in QP, dividing by P and then adding the plaintext.
+	// The level of the output ciphertext is plaintext.Level().
+	EncryptNew(plaintext *Plaintext) *Element
+
+	// EncryptNTTNew encrypts the input plaintext using the stored key and returns
+	// the result on a newly created ciphertext. The encryption is done by first
+	// encrypting zero in QP, dividing by P and then adding the plaintext.
+	// The level of the output ciphertext is plaintext.Level().
+	// Output will be in the NTT domain.
+	EncryptNTTNew(plaintext *Plaintext) *Element
```
ChristianMct, line 24: Same comment as for the Decryptor, if these NTT variants are for CKKS, I would try to support one type of encryption at rlwe level and add a final NTT at the ckks level (if doable and it has not a too big impact on perf of course).

rlwe/encryptor.go
```diff
@@ -0,0 +1,509 @@
...
+	// encrypting zero in QP, dividing by P and then adding the plaintext.
+	// The level of the output ciphertext is min(plaintext.Level(), ciphertext.Level()).
+	// Output domain will match ciphertext.Value[0].IsNTT value.
+	Encrypt(plaintext *Plaintext, ciphertext *Element)
+
+	// EncryptFastNew encrypts the input plaintext using the stored key and returns
+	// the result on a newly created ciphertext. The encryption is done by first
+	// encrypting zero in Q and then adding the plaintext.
+	// The level of the output ciphertext is plaintext.Level().
+	EncryptFastNew(plaintext *Plaintext) *Element
```
ChristianMct, line 37: As discussed some time ago, one way to reduce this interface would be to have two implementation of the encryptor, one for each encryption algo. Then the user chooses the type of encryption at initialization with something like: `NewEncryptor`  and `NewFastEncryptor`.

bfv/encryptor.go
Pro7ech: fixed

ckks/encryptor.go
Pro7ech: fixed

bfv/decryptor.go
Pro7ech: fixed

bfv/keygen.go
Pro7ech: fixed

ckks/decryptor.go
Pro7ech: fixed

ckks/encryptor.go
Pro7ech: fixed

bfv/keygen.go
```diff
@@ -1,254 +1,36 @@
...
-	GenRotationKeysForRotations(ks []int, includeSwapRow bool, sk *rlwe.SecretKey) (rks *rlwe.RotationKeySet)
+	GenSwitchingKeyForRotationBy(k int, sk *rlwe.SecretKey) (swk *rlwe.SwitchingKey)
+	GenRotationKeysForRotations(ks []int, inclueSwapRows bool, sk *rlwe.SecretKey) (rks *rlwe.RotationKeySet)
+	GenSwitchingKeyForRowRotation(sk *rlwe.SecretKey) (swk *rlwe.SwitchingKey)
 	GenRotationKeysForInnerSum(sk *rlwe.SecretKey) (rks *rlwe.RotationKeySet)
 }
 
 // keyGenerator is a structure that stores the elements required to create new keys,
 // as well as a small memory pool for intermediate values.
 type keyGenerator struct {
```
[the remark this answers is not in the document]
Pro7ech: fixed

ckks/keygen.go
Pro7ech: Modification in the evaluator would need to be done to accommodate for arbitrary degree tensoring and relinearization

rlwe/decryptor.go
Pro7ech: Because the plaintext is created by the decryptor, it must be told what the output domain of the plaintext is.

rlwe/encryptor.go
Pro7ech: fixed

rlwe/encryptor.go
Pro7ech: Same, the decryptor creates the plaintext, so it must be told what the output domain is.

rlwe/encryptor.go
Pro7ech: I would rather keep a general interface and have more methods than split it into two different interfaces

rlwe/decryptor.go
Pro7ech: Remove [...]New methods from RLWE and wrap inplace method in schemes

rlwe/encryptor.go
Pro7ech: Remove [...]New methods from RLWE and wrap inplace method in schemes

rlwe/encryptor.go
Pro7ech: Ok, then split both and have the scheme encryptor constructor instantiate the correct one based on the provided parameters

rlwe/decryptor.go
ChristianMct: Done

rlwe/encryptor.go
ChristianMct: Done

rlwe/encryptor.go
ChristianMct: Done

rlwe/decryptor.go
ChristianMct: done

PR #153, opened 2021-11-04
rlwe ciphertext marshalling

Pro7ech:


bfv/marshaler.go
```diff
@@ -1,66 +1,21 @@
 package bfv
```
ChristianMct, line 1: Since marshalers are well modularized now, I think there is no point in keeping `bfv/marshaler.go`.  The methods could be moved to `bfv/ciphertext.go`.

ckks/marshaler.go
```diff
@@ -3,24 +3,19 @@ package ckks
 import (
```
ChristianMct, line 3: Since marshalers are well modularized now, I think there is no point in keeping `ckks/marshaler.go`.  The methods could be moved to `ckks/ciphertext.go`.

rlwe/rlwe_test.go
```diff
@@ -537,6 +538,44 @@ func testMarshaller(kgen KeyGenerator, t *testing.T) {
...
+
+		require.Equal(t, ciphertextWant.Degree(), ciphertextTest.Degree())
+		require.Equal(t, ciphertextWant.Level(), ciphertextTest.Level())
+
+		for i := range ciphertextWant.Value {
+			require.True(t, params.RingQ().EqualLvl(ciphertextWant.Level(), ciphertextWant.Value[i], ciphertextTest.Value[i]))
+		}
+	})
+
+	t.Run(testString(params, "Marshaller/Ciphertext/Minimal/"), func(t *testing.T) {
```
ChristianMct, line 561: I don't get the point of separating in two tests cases? Can the two be merged in one, potentially with an internal loop repeating the test for several ct degree ?.

rlwe/elements.go
```diff
@@ -98,6 +98,13 @@ func NewCiphertextNTT(params Parameters, degree, level int) *Ciphertext {
 	return el
 }
 
+// NewCiphertextRandom generates a new uniformly distributed Ciphertext of degree, level and scale.
+func NewCiphertextRandom(prng utils.PRNG, params Parameters, degree, level int) (ciphertext *Ciphertext) {
+	ciphertext = NewCiphertext(params, degree, level)
+	PopulateElementRandom(prng, params, ciphertext)
```
ChristianMct, line 104: This method creates a new Ring from the passed params which is no longer needed.

ChristianMct (changes requested):

rlwe/elements.go
Pro7ech: Fixed

rlwe/rlwe_test.go
Pro7ech: I suppose, only one test with degree 2 would be sufficient

ckks/marshaler.go
Pro7ech: fixed

bfv/marshaler.go
Pro7ech: fixed

rlwe/rlwe_test.go
ChristianMct: Added the test for degree 0, 1, 2, 3, 4 for future-proofness.

ChristianMct (approved):

PR #165, opened 2022-01-09
[ckks/bootstrapping] : Improve bootstrapping

Pro7ech:


n-samar:
Can you please merge this branch in? Awesome work btw; thank you!

ChristianMct:
Yes, the plan is to merge it. We're closing on some other more urgent changes right now, but this one will be merge right after these are complete.

examples/ckks/bootstrapping/main.go
```diff
@@ -24,20 +24,23 @@ func main() {
 	var plaintext *ckks.Plaintext
 
 	// Bootstrapping parameters
-	// Four sets of parameters (index 0 to 3) ensuring 128 bit of security
-	// are available in github.com/tuneinsight/lattigo/v3/ckks/bootstrap_params
+	// 2xFour sets of parameters (index 0 to 3) ensuring 128 bit of security
```
ChristianMct, line 27: Eight, ... index 0 to 7.

examples/ckks/bootstrapping/main.go
```diff
@@ -24,20 +24,23 @@ func main() {
...
+	// are available in github.com/ldsec/lattigo/v2/ckks/bootstrapping/default_params.go
```
ChristianMct, line 28: `github.com/tuneinsight/lattigo/v3/ckks/bootstrapping/default_params.go`

ckks/bootstrapping/bootstrapper.go
```diff
@@ -28,10 +28,21 @@ type bootstrapperBase struct {
...
 
 	q0OverMessageRatio float64
+
+	swkDtS *rlwe.SwitchingKey
+	swkStD *rlwe.SwitchingKey
+}
+
+// Key is a type for a CKKS bootstrapping key, wich regroups the necessary public relinearization
+// and rotation keys (i.e., an EvaluationKey).
+type Key struct {
```
ChristianMct, line 38: For consistency, I would change this type slightly to disambiguate the term "Key", which could refer to a secret.
For example:

```suggestion
type EvaluationKey struct {
    *rlwe.EvaluationKey
     SwkDtS *rlwe.SwitchingKey
     SwkStD *rlwe.SwitchingKey
}
```

This is still coherent with the package name: `bootstrapping.EvaluationKey`.

examples/ckks/bootstrapping/main.go
```diff
@@ -53,7 +56,8 @@ func main() {
 	rotations := btpParams.RotationsForBootstrapping(params.LogN(), params.LogSlots())
 	rotkeys := kgen.GenRotationKeysForRotations(rotations, true, sk)
 	rlk := kgen.GenRelinearizationKey(sk, 2)
```
ChristianMct, line 58: Should we add a `KeyGen` type and a `GenEvaluationKeyForBootstrapping` function for convenience ?

ChristianMct:
I think it might be valuable to keep the possibility of doing bootstrapping without secret-key encapsulation for backward compatibility. A zero value for the ephemeral secret-key could be a condition for skipping the keyswitch in the modup ?

examples/ckks/bootstrapping/main.go
Pro7ech: The two set of four parameters are in different structs.

ckks/bootstrapping/bootstrapper.go
Pro7ech: works for me, although `bootstrapping.Key` should already reflect the purpose of the key.

ChristianMct: But doesn't help figuring out that this key is not private. With the suggestion, we are explicit that this object is an evaluation key. We could add "Public" somewhere but we didn't for the rest of the library.

examples/ckks/bootstrapping/main.go
ChristianMct: Or simply remove the `github.com/ldsec/lattigo/v2/` prefix to avoid having to update it at each major version.

examples/ckks/bootstrapping/main.go
ChristianMct: I see. Then it should be better explained that four sparse-key sets and four dense-key sets are available at...

n-samar:
Thank you very much for working on merging this and being so responsive. I really appreciate it!

examples/ckks/bootstrapping/main.go
Pro7ech: Done

jrtroncoso (approved):

PR #166, opened 2022-01-24
Add support for secret-key parameterization from hamming weight

ChristianMct:


rlwe/params.go
```diff
@@ -88,14 +91,20 @@ func NewParameters(logn int, q, p []uint64, sigma float64, ringType ring.Type) (
 // NewParametersFromLiteral instantiate a set of generic RLWE parameters from a ParametersLiteral specification.
 // It returns the empty parameters Parameters{} and a non-nil error if the specified parameters are invalid.
 func NewParametersFromLiteral(paramDef ParametersLiteral) (Parameters, error) {
+
```
Pro7ech, line 94: We should put somewhere what fields are mandatory or can be left empty, and what happens if done so

rlwe/params.go
```diff
@@ -357,7 +372,11 @@ func (p Parameters) CopyNew() Parameters {
 	qi, pi := p.qi, p.pi
 	p.qi, p.pi = make([]uint64, len(p.qi)), make([]uint64, len(p.pi))
 	copy(p.qi, qi)
-	copy(p.pi, pi)
+	p.ringQ, _ = ring.NewRingFromType(1<<p.logN, p.qi, p.ringType)
```
Pro7ech, line 375: I think we should remove this method since parameters are now immutable and read only

rlwe/rlwe_test.go
```diff
@@ -136,11 +136,42 @@ func testGenKeyPair(kgen KeyGenerator, t *testing.T) {
 
 	params := kgen.(*keyGenerator).params
 
+	sk, pk := kgen.GenKeyPair()
+
+	// Checks that the secret-key has exactly params.h non-zero coefficients
+	t.Run(testString(params, "SK"), func(t *testing.T) {
```
Pro7ech, line 142: Isn't this already tested in the package ring?

ChristianMct: Do you mean as a part of the `TernarySampler` tests ? 

We need to test it at that layer anyway, to catch bugs at the `KeyGen` level (e.g., using the wrong sampler).

rlwe/params.go
ChristianMct: Agreed.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 - RING: `.PolyToBigint[...](*)` now take as input `gap` which defines the multiples of `X^{i*gap}` to reconstruct.
 - RLWE: removed `FastEncryptor`. Encryption without rescaling by `P` is now automatically used by `Encryptor` if no `P` is specified in the parameters.
 - RLWE: `NewAdditiveShareBigint` now takes as input the size of the share.
+- RLWE: added the `H` field and `HammingWeight` method for parameters-related structs to specify and determing the secret-key distribtion.
```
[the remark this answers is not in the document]
ChristianMct: Agreed, all secrets are now sampled according to H, not only the secret-keys.

rlwe/params.go
ChristianMct: Yes. If we agree with the design, I'll make a pass on the documentation to reflect that.

rlwe/rlwe_test.go
Pro7ech: I'm agree

CHANGELOG.md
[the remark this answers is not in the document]
ChristianMct: Done.

rlwe/params.go
ChristianMct: Done.

CHANGELOG.md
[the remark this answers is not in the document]
ChristianMct: Done.

rlwe/params.go
ChristianMct: For now, I only deprecated the method:
- to not make the code that might use it crash
- because `Parameter` still has one mutable method (`UnmarshalBinary`)

jrtroncoso (approved):

PR #171, opened 2022-02-09
fixed issue 170

Pro7ech:
DRLWE allows the generation of public-keys and key-switch with parameters without the auxiliary modulus P.


jrtroncoso (changes requested):
Question: Should the NewUniformSamplerQP panic if both RingQ and RingP are nil?

jrtroncoso (approved):

PR #180, opened 2022-02-22
[examples] added dckks examples

wonderit:
Added dckks/psi, dckks/pir examples based on CKKS.


ChristianMct:
Hi @wonderit,

Thanks for sharing your example. Indeed, we don't have examples for the `dckks` package. Yet, we would like these examples to be different from the `dbfv` ones, in order to better illustrate the particularities and functionalities of schemes. 

From your experience in implementing `dckks/psi` and `dckks/pir`, which of BFV or CKKS is the best fit for the usecase ?

wonderit:
Hi @ChristianMct, 

Thanks for the feedback. I agree with the direction of trying to make examples for the `dckks` package different from `dbfv`. I guess the features and functionalities of ckks schemes can be better illustrated by implementing examples such as logistic regression using the approximation of a sigmoid function or training neural network using iterative multiplication.

While implementing the `dckks` example, I thought that `dckks/psi` is an example to generate collective relinearization key and `dckks/pir` is to generate collective rotation key. Since `psi` and `pir` are examples that receive integer indexes as inputs, BFV seems to be suitable. However, if the input changes to float and aims for `vector multiplication`(instead of `psi`) or `dot product`(instead of `pir`), CKKS seems to be suitable.

ChristianMct:
>  I agree with the direction of trying to make examples for the dckks package different from dbfv

We will be on the lookout for contributions in that direction. 

>  I thought that dckks/psi is an example to generate collective relinearization key and dckks/pir is to generate collective rotation key

The MHE setup phase is in fact independent of the encryption scheme. Relinearization- and rotation-keys generated with the `drlwe` package can be used for both CKKS and BFV. Hence, examples for how they are generated belong should be at the `drlwe` level. I have started #283 to implement these changes, and will therefore close this pull request.

PR #190, opened 2022-03-18
[ckks]: fixed mul&add correctness for non-identical inputs

Pro7ech:
- Fixed `MulAndAdd` correctness when `op0 != op1`.
- Added test for  `op0 != op1`.

### DIFF
--- a/ckks/ckks_test.go
+++ b/ckks/ckks_test.go
@@ -713,11 +713,30 @@ func testEvaluatorMulAndAdd(tc *testContext, t *testing.T) {
 
 		tc.evaluator.MulRelinAndAdd(ciphertext2, ciphertext2, ciphertext1)
 
-		//fmt.Println(ciphertext1.Value[2].Coeffs[0][:4])
-
 		require.Equal(t, ciphertext1.Degree(), 1)
 
 		verifyTestVectors(tc.params, tc.encoder, tc.decryptor, values1, ciphertext1, tc.params.LogSlots(), 0, t)
+	})
+
+	t.Run(GetTestName(tc.params, "Evaluator/MulAndAdd/ct0*ct1->ct2"), func(t *testing.T) {
+
+		values1, _, ciphertext1 := newTestVectors(tc, tc.encryptorSk, complex(-1, -1), complex(1, 1), t)
+		values2, _, ciphertext2 := newTestVectors(tc, tc.encryptorSk, complex(-1, -1), complex(1, 1), t)
+
+		for i := range values1 {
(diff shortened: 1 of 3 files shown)


ckks/evaluator.go
```diff
@@ -1278,6 +1280,10 @@ func (eval *evaluator) mulRelinAndAdd(op0, op1 Operand, relin bool, ctOut *Ciphe
 		panic("cannot MulRelinAndAdd: input elements must be of degree 0 or 1")
 	}
 
+	if op0.El() == ctOut.El() || op1.El() == ctOut.El() {
+		panic("ctOut must be different fron op0 and op1")
```
ChristianMct, line 1284: typo

ckks/evaluator.go
```diff
@@ -1260,6 +1261,7 @@ func (eval *evaluator) MulAndAdd(op0, op1 Operand, ctOut *Ciphertext) {
 // The procedure will panic if either op0.Degree or op1.Degree > 1.
 // The procedure will panic if ctOut.Degree != op0.Degree + op1.Degree.
 // The procedure will panic if the evaluator was not created with an relinearization key.
+// The procedure will panic if ctOut = op0 or op1.
```
ChristianMct, line 1264: Is it really necessary ? For MulRelinAndAdd the inputs and outputs ciphertexts could be of the same degree.

ckks/ckks_test.go
```diff
@@ -713,13 +713,32 @@ func testEvaluatorMulAndAdd(tc *testContext, t *testing.T) {
...
 		tc.evaluator.MulRelinAndAdd(ciphertext2, ciphertext2, ciphertext1)
 
-		//fmt.Println(ciphertext1.Value[2].Coeffs[0][:4])
-
 		require.Equal(t, ciphertext1.Degree(), 1)
 
 		verifyTestVectors(tc.params, tc.encoder, tc.decryptor, values1, ciphertext1, tc.params.LogSlots(), 0, t)
 	})
 
+	t.Run(GetTestName(tc.params, "Evaluator/MulAndAdd/ct0*ct1->ct2"), func(t *testing.T) {
```
ChristianMct, line 721: Add the same test for MulRelinAndAdd ?

Pro7ech: It is not necessary, the relinearization is a second step, independent from what is happening internally before it, so it only needs to be tested once as it is not affected by the inputs.

ckks/evaluator.go
Pro7ech: Fixed

ChristianMct (approved):

PR #212, opened 2022-05-25
Dev dckks params switch

Pro7ech:
Upgrade DCKKS MaskedTransform to enable params switch (different degree, modulus and slots).


Pro7ech:
Feature has been included in V4.0.0

PR #222, opened 2022-07-22
Improvement to polynomial evaluation and linear transformations

Pro7ech:
- Generalization of polynomials: one interface bases struct to rule them all
- Plaintext, Encoded and Encrypted polynomials, with marshalling.
- Pre-compute the baby-step giant-step format of polynomials (with scaled coefficients)
- Encrypted linear transformations
- Marshaling for linear transformations


Pro7ech:
functionalities will be extracted to a new PR

PR #283, opened 2022-09-23
Improve drlwe example and documentation

ChristianMct:
The goal is to improve the examples and documentation of the `drlwe` package to make it clear that the MHE setup phase that it implement can be used for all schemes.

This removes the need to duplicate examples between the `dbfv` and `dckks` as pointed out in #180.


Pro7ech:
This branch was merged in main through update_README

PR #285, opened 2022-10-04
CI: add Go version 1.19 to the matrix of tests

romainbou:
It also updates `staticcheck` to `1.19`


romainbou:
With Go 1.19. `go fmt` was updated to format the go doc which would unfortunately break some of the ascii diagrams.
e.g.
Screenshot from 2022-10-04 12-26-35

So I fixed them using "2 spaces" for markdown to interpret them as code

Pro7ech (approved):

rlwe/params.go
```diff
@@ -478,7 +478,8 @@ func (p Parameters) GaloisElementsForMergeRLWE() (galEls []uint64) {
 }
 
 // InverseGaloisElement takes a Galois element and returns the Galois element
-//  corresponding to the inverse automorphism
+//
+//	corresponding to the inverse automorphism
```
ChristianMct, line 482: was this one introduced by mistake ?

romainbou: Yes, there were two spaces, so `go fmt` re-indented it automatically. In this case I'll just remove the extra space, it fixes it.

On the other hand, I notice that in other instances where there was intentional indentation for in a list of value and it automatically created those extra new lines which I think are confusing.
https://github.com/tuneinsight/lattigo/blob/24f14142881cdc49306cf22291e1b33f91dc2694/ckks/encoder.go#L323-L335

To maintain indentation, I can't find a better solution than just adding an extra character (e.g. `>`) at the line beginning and then indent: e.g. 
1.
```go
// This method it as the core of the slot encoding.
// values: values.(type) can be either []complex128 of []float64.
// >       The imaginary part of []complex128 will be discarded if ringType == ring.ConjugateInvariant.
// logslots: user must ensure that 1 <= len(values) <= 2^logSlots < 2^logN and that logSlots >= 3.
```

2. or simply not indent:
```go
// This method it as the core of the slot encoding.
// values: values.(type) can be either []complex128 of []float64.
// The imaginary part of []complex128 will be discarded if ringType == ring.ConjugateInvariant.
// logslots: user must ensure that 1 <= len(values) <= 2^logSlots < 2^logN and that logSlots >= 3.
```

3. or to actually use a list for each parameter (but them it'd be good to consistently do the same when elsewhere):
```go
// This method it as the core of the slot encoding.
//   - values: values.(type) can be either []complex128 of []float64.
//     The imaginary part of []complex128 will be discarded if ringType == ring.ConjugateInvariant.
//   - logslots: user must ensure that 1 <= len(values) <= 2^logSlots < 2^logN and that logSlots >= 3.
//   - scale: the scaling factor used do discretize float64 to fixed point integers.
//   - montgomery: if true then the value written on polyOut are put in the Montgomery domain.
//   - polyOut: polyOut.(type) can be either ringqp.Poly or *ring.Poly.
//     The encoding encoding is done at the level of polyOut.
//     Values written on  polyOut are always in the NTT domain.
```

Let me know what you prefer :wink:

ChristianMct: Since Go 1.19 should recognize the lists in the doc comments, I'd go for the last option.

This will be less readable when browsing the doc from a < 1.19 rendering, but it's not a major issue in my opinion.

ChristianMct (approved):

PR #296, opened 2022-11-10
Simplify and generalize interface to `defaultParametersLiteral`

n-samar:
Introduce `FrontendParameters`, a reduced description of CKKS parameters that abstracts away implementation details in `defaultParameterLiteral` and allows for arbitrary combinations of `LogN`, `LogScale`, `UsableLevels`, and `BoostrappingPrecision`.

### DIFF
--- a/ckks/bootstrapping/bootstrapping_bench_test.go
+++ b/ckks/bootstrapping/bootstrapping_bench_test.go
@@ -13,7 +13,7 @@ func BenchmarkBootstrap(b *testing.B) {
 	var err error
 	var btp *Bootstrapper
 
-	paramSet := DefaultParametersDense[0]
+	paramSet := MakeFrontendParameters(16, 40, 10, 26).GetDefaultParametersSparse()
 	ckksParams := paramSet.SchemeParams
 	btpParams := paramSet.BootstrappingParams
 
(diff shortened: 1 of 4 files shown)


Pro7ech:
Hi @n-samar, thank you for your interest in Lattigo. I very well understand the need to have a better abstraction of the bootstrapping parameters to increase their usability and I have been thinking about how to do it for a while now.

We currently have the open #255 aimed at refactoring the usability of the bootstrapping. Its work has been delayed because of the current refactoring of #275 that notably changes how the library manages the scales, which is API breaking. The good news is that #275 is currently under review and it should be merged in the coming days and this will allow me to restart the work on #255. 

#255 is only at its early stage and many features are planned for it, such as:
- Bootstrapping interface
- Separate the bootstrapping parameters from the CKKS parameters and only specify the high level description of the bootstrapping, see this [file ](https://github.com/tuneinsight/lattigo/blob/dev_bootstrapping/ckks/bootstrapping/default_params.go)
- Tools to enable the bootstrapping one or more small ciphertexts in a larger one with a single bootstrapping
- Tools to enable the bootstrapping for the Conjugate-Invariant variant of CKKS
- And more

n-samar:
Great! Thanks so much for a great library and prompt response. We couldn't do a lot of our research without Lattigo. Please do let me know when #255 is ready for review.

Pro7ech:
> 

Hi @n-samar, #297 is ready for review. It still misses a few functionalities that I would like to include, but it should have all the ones that you were waiting for, so you can take a look and test it. Don't hesitate to give feedback and to ask for something that you think is missing, this is the time!

Pro7ech:
Closing this pull request because the request functionalities are provided by #255 (which was then reopened as #297).

PR #297, opened 2022-11-17
Dev bootstrapping 2.0

Pro7ech:
c.f. CHANGELOG.md


ckks/bootstrapping/bootstrapping.go
```diff
@@ -8,47 +8,75 @@ import (
 	"github.com/tuneinsight/lattigo/v4/rlwe"
 )
 
-// Bootstrap re-encrypts a ciphertext at lvl Q0 to a ciphertext at MaxLevel-k where k is the depth of the bootstrapping circuit.
+// Bootstrapp re-encrypt a ciphertext at lvl Q0 to a ciphertext at MaxLevel-k where k is the depth of the bootstrapping circuit.
 // If the input ciphertext level is zero, the input scale must be an exact power of two smaller or equal to round(Q0/2^{10}).
 // If the input ciphertext is at level one or more, the input scale does not need to be an exact power of two as one level
 // can be used to do a scale matching.
-func (btp *Bootstrapper) Bootstrap(ctIn *rlwe.Ciphertext) (ctOut *rlwe.Ciphertext) {
+func (btp *Bootstrapper) Bootstrapp(ctIn *rlwe.Ciphertext) (ctOut *rlwe.Ciphertext) {
```
n-samar, line 16: In `examples/ckks/bootstrapping/main.go`, there is a comment:
```
// CAUTION: the scale of the ciphertext MUST be equal (or very close) to params.DefaultScale()
```
I _strongly_ advise that you assert this condition here explicitly (panicking if the condition is not met), at the top of `Bootstrapp()` and remove the CAUTION comment. I lost ~3 days tracing down a bug that would have been obvious if there was an assertion here.

examples/ckks/bootstrapping/main.go
```diff
@@ -95,7 +120,7 @@ func main() {
 	// To equalize the scale, the function evaluator.SetScale(ciphertext, parameters.DefaultScale()) can be used at the expense of one level.
```
n-samar, line 120: Please see my comment on the `Bootstrapp()` function. The condition documented in this comment should be checked in code.

ckks/params.go
```diff
@@ -311,7 +292,7 @@ func (p ParametersLiteral) RLWEParameters() rlwe.ParametersLiteral {
 		H:              p.H,
 		RingType:       p.RingType,
 		DefaultNTTFlag: DefaultNTTFlag,
-		DefaultScale:   rlwe.NewScale(p.DefaultScale),
+		DefaultScale:   rlwe.NewScale(math.Exp2(float64(p.LogScale))),
```
n-samar, line 295: Both `float64(p.LogScale)` and `math.Exp2(float64)` can return incorrect results silently. I was unable to find nice golang standard library stuff that `panic()`s instead of returning incorrect results silently, so if you know them, please use them. An alternative is to implement a function:
```
panic_float64(value: int64) {
  assert(value < (1 << 52))
  return float64(value)
}
```
and
```
fn panic_exp2(value: float64) {
  let result = math.Exp2(value)
  assert(!math.IsNan(result) and !math.IsInf(result) and result != 0.0)
  return result
}
```
(I don't know Go, but you get the picture).

Then do 
```
rlwe.NewScale(panic_exp2(panic_float64(p.LogScale))
```

While these edgecases are unlikely, if they were ever to happen, tracking down these bugs would be very hard.

ckks/params.go
```diff
@@ -393,15 +374,15 @@ func (p Parameters) StandardParameters() (pckks Parameters, err error) {
...
-		DefaultScale: p.DefaultScale().Float64(),
-		LogSlots:     p.LogSlots(),
+		LogN:     p.LogN(),
+		Q:        p.Q(),
+		P:        p.P(),
+		Pow2Base: p.Pow2Base(),
+		Sigma:    p.Sigma(),
+		H:        p.HammingWeight(),
+		RingType: p.RingType(),
+		LogScale: int(math.Round(math.Log2(p.DefaultScale().Float64()))),
```
n-samar, line 384: LogScale should probably be an `int64`, not an `int`. Otherwise, this operation may overflow on 32-bit systems.

ckks/bootstrapping/default_params.go
```diff
@@ -24,674 +22,174 @@ var DefaultParametersDense = []defaultParametersLiteral{N16QP1767H32768H32, N16Q
...
-					{0xfffffffff00001},
-					{0xffffffffd80001},
-					{0x1000000002a0001},
-				},
-			},
+			LogN:     16,
+			LogQ:     []int{60, 40, 40, 40, 40, 40, 40, 40, 40, 40},
+			LogP:     []int{61, 61, 61, 61, 61},
+			H:        192,
+			LogScale: 40,
```
n-samar, line 37: Is `LogScale` ever not equal to the small moduli in `LogQ` (excluding the first one)? I don't think being able to create these "raw" `ParametersLiteral` like this is safe because type invariants are not checked. Please see my comment on using a "constructor" above.

ckks/bootstrapping/default_params.go
```diff
@@ -24,674 +22,174 @@ var DefaultParametersDense = []defaultParametersLiteral{N16QP1767H32768H32, N16Q
...
+			LogP:     []int{61, 61, 61, 61},
+			H:        192,
+			LogScale: 45,
+		},
+		ParametersLiteral{
+			S2CLogScale:     [][]int{{42}, {42}, {42}},
+			C2SLogScale:     [][]int{{58}, {58}, {58}, {58}},
+			LogMessageRatio: 2,
+			ArcSineDeg:      7,
+			EvalModLogScale: 60,
```
n-samar, line 62: These `LogScale`s probably have to be bigger than that in `LogScale`? If yes, I think a "constructor" should be use and this invariant should be explicitly checked.

n-samar (commented):
Thank you for your work and giving me an opportunity to comment!

I hope my comments do not seem harsh: my primary goal was to comment clearly and concisely.

The bootstrapping interface is much more convenient for me the way it is now. Thank you! Please see the code comments for details and suggestions.

Nikola

ckks/bootstrapping/bootstrapping.go
```diff
@@ -8,47 +8,75 @@ import (
...
```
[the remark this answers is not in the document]
Pro7ech: My mistake, this is a typo, I will correct that

ckks/bootstrapping/bootstrapping.go
```diff
@@ -8,47 +8,75 @@ import (
...
+		btp.Rescale(tmp, btp.params.DefaultScale(), tmp)
+
+		// [2^d * M + 2^(d-2n) * e'] <- [2^d * M + 2^(d-n) * e] - [2^(d-n) * e + 2^(d-2n) * e']
+		btp.Add(ctOut, tmp, ctOut)
 	}
 
+	return
+}
+
+func (btp *Bootstrapper) bootstrapp(ctIn *rlwe.Ciphertext) (ctOut *rlwe.Ciphertext) {
```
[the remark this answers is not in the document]
Pro7ech: same

examples/ckks/bootstrapping/main.go
```diff
@@ -11,63 +11,88 @@ import (
...
+	// to the constructor along with the specificities of the bootstrapping circuit we choose, to
+	// enable it to create the appropriate ckks.ParametersLiteral that enable the evaluation of the
+	// bootstrapping circuit on top of the residual moduli that we defined.
+	ckksParamsLit := ckks.ParametersLiteral{
+		LogN:     16,                                                // Log2 of the ringdegree
+		LogSlots: 15,                                                // Log2 of the number of slots
+		LogQ:     []int{55, 40, 40, 40, 40, 40, 40, 40, 40, 40, 40}, // Log2 of the ciphertext prime moduli
+		LogP:     []int{61, 61, 61, 61},                             // Log2 of the key-switch auxiliary prime moduli
+		LogScale: 40,                                                // Log2 of the scale
+		H:        192,                                               // Hamming weight of the secret
```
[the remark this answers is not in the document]
Pro7ech: This how Lattigo is designed: `ckks.ParametersLiteral` allows the user to create raw parameters, and then a constructor takes this struct, checks them and produces an unmodifiable & read only `ckks.Parameters` struct.

ckks/bootstrapping/bootstrapping.go
Pro7ech: Thank you for pointing it out, I added this check.

ckks/params.go
Pro7ech: The default scaling factor cannot be given as a float anymore. It is now given in log2 as an int, which means that it will always be a power of two, and thus will always have an exact representation. As the default scaling factor is read only, it can never be modified, such edge cases can never happen.

ckks/params.go
Pro7ech: By design, all parameters and variables in Lattigo that do not interact with polynomials coefficients (which are type `uint64`) use the default type which is `int`.

ckks/bootstrapping/default_params.go
Pro7ech: `ParametersLiteral` are checked by the constructor `NewParametersFromLiteral`. We want to keep the maximum possible flexibility for the user uppon defining the `ParametersLiteral`, such as being able to specify the size, or even the prime, of the moduli chain. This is required for circuit optimization.

ckks/bootstrapping/default_params.go
Pro7ech: These are bootstrapping parameters (homomorphic encoding and decoding), which are independent of the scaling factor. They provide verified performances that are detailed in the godoc. Advanced users can specify different values to customize the bootstrapping circuit.

examples/ckks/bootstrapping/main.go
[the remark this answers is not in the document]
n-samar: Nice! That makes sense.

I've always been confused by why there are both `ParametersLiteral` and `Parameters`. Looking at their definitions, it looks like the distinction is documented well though. I just never read it. Thanks!

ckks/bootstrapping/default_params.go
n-samar: This makes sense. Thank you!

ckks/bootstrapping/default_params.go
n-samar: Got it. Thank you!

ckks/params.go
n-samar: OK. I suggest that we verify that `LogScale` doesn't overflow then by using a function:
```
fn panic_log2(value: float64) -> int {
  // Ensure `log2(value)` is smaller than 2^(32)-1
  assert(value < math.Exp2(float64((int64(1) << 32) - 1)))
  return math.Log2(value)
}
```

ckks/params.go
```diff
@@ -285,17 +266,17 @@ var (
 // type (RingType) and the number of slots (in log_2, LogSlots). If left unset, standard default values for
 // these field are substituted at parameter creation (see NewParametersFromLiteral).
 type ParametersLiteral struct {
```
n-samar, line 268: The comment above says:
```
// ParametersLiteral is a literal representation of BFV parameters
```
Shouldn't it say _CKKS_ parameters instead? Sorry for mentioning this when it's not related to the current PR, but I will forget otherwise and it's probably worth fixing if it is a mistake.

ckks/params.go
Pro7ech: I don't this this can ever happen because `2^{2^{32}} = 2^{4294967296}` isn't a number that can be represented, it will return `+Inf`. I have thus added a check in `rlwe.Scale` that will panic if the scaling factor is `Inf` (`NaN` value will by default trigger a panic if used to instantiate a `*big.Float`).

ckks/params.go
Pro7ech: Yes this is a typo, that I also just noticed, thank you.

ckks/params.go
n-samar: `2^{2^32}` can be represented in `float64` ([see here](https://en.wikipedia.org/wiki/Double-precision_floating-point_format)).

n-samar: Actually no. I was wrong! Sorry!

ckks/bootstrapping/bootstrapping.go
```diff
@@ -8,47 +8,79 @@ import (
 	"github.com/tuneinsight/lattigo/v4/rlwe"
 )
 
-// Bootstrap re-encrypts a ciphertext at lvl Q0 to a ciphertext at MaxLevel-k where k is the depth of the bootstrapping circuit.
+// Bootstrap re-encrypt a ciphertext at lvl Q0 to a ciphertext at MaxLevel-k where k is the depth of the bootstrapping circuit.
 // If the input ciphertext level is zero, the input scale must be an exact power of two smaller or equal to round(Q0/2^{10}).
```
n-samar, line 12: This condition should probably be explicitly checked in the function (and then the comment should be removed). I am not sure what 2^10 means here exactly? It's probably assumed `2^{10} > ||m||^{canonical}_{\infty}` or something like that?

Pro7ech: Yes you are correct, and it is also not clear. I will update the godoc of the method.

ckks/advanced/marshaler.go
```diff
@@ -5,84 +5,140 @@ import (
...
+	dataLen += len(mParams.Levels) // Levels
+	dataLen++                      // RepackImag2Real
+	dataLen += 8                   // Scaling
+	dataLen++                      // BitReversed
+	dataLen++                      // LogBSGSRatio
+	return
+}
+
 // MarshalBinary encode the target EncodingMatrixParameters on a slice of bytes.
 func (mParams *EncodingMatrixLiteral) MarshalBinary() (data []byte, err error) {
```
ChristianMct, line 23: I'd strongly advise against "optimal" binary marshalers for literal structs:
- they have to be adapted manually
- params serialization is usually not a performance bottleneck
- JSON serializers are usually much more user-friendly and composable
- `MarshalBinary()` can be `return mParams.MarshalJSON()`

ckks/bootstrapping/bootstrapping.go
```diff
@@ -2,53 +2,85 @@
...
+	"fmt"
 	"math"
 
 	"github.com/tuneinsight/lattigo/v4/ring"
 	"github.com/tuneinsight/lattigo/v4/rlwe"
 )
 
-// Bootstrap re-encrypts a ciphertext at lvl Q0 to a ciphertext at MaxLevel-k where k is the depth of the bootstrapping circuit.
-// If the input ciphertext level is zero, the input scale must be an exact power of two smaller or equal to round(Q0/2^{10}).
+// Bootstrap re-encrypts a ciphertext to a ciphertext at MaxLevel-k where k is the depth of the bootstrapping circuit.
```
ChristianMct, line 12: `MaxLevel - k ` (spacing so no confusion with a dash)

ckks/bootstrapping/bootstrapping.go
```diff
@@ -2,53 +2,85 @@
...
+// If the input ciphertext level is zero, the input scale must be an exact power of two smaller than Q[0]/MessageRatio
+// (it can't be equal since Q[0] is not a power of two).
+// The message ratio is an optional field in the bootstrapping parameters, by default it set to 2^{LogMessageRatio = 8}.
```
ChristianMct, line 15: few words on what is the message ratio.

ckks/bootstrapping/dummy_bootstrapping.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"fmt"
+
+	"github.com/tuneinsight/lattigo/v4/ckks"
+	"github.com/tuneinsight/lattigo/v4/rlwe"
+)
+
+// DummyBootstrapper is a dummy bootstrapper that complies to
+// the rlwe.Bootstrapping interface
+type DummyBootstrapper struct {
```
ChristianMct, line 12: DummyBootstrapper is an implementation of the rlwe.Bootstrapping interface that uses the secret-key to decrypt and re-encrypt the bootstrapped ciphertext.

ckks/bootstrapping/dummy_bootstrapping.go
```diff
@@ -0,0 +1,49 @@
+package bootstrapping
+
+import (
+	"fmt"
+
+	"github.com/tuneinsight/lattigo/v4/ckks"
+	"github.com/tuneinsight/lattigo/v4/rlwe"
+)
+
+// DummyBootstrapper is a dummy bootstrapper that complies to
```
ChristianMct, line 10: Type name might be confusing:
- does not explain why it is "dummy"
- Could be useful outside testing

ckks/bootstrapping/parameters.go
```diff
@@ -0,0 +1,635 @@
...
+	Iterations            int               // Default:  1
+	SineType              advanced.SineType // Default: advanced.CosDiscret
+	LogMessageRatio       int               // Default: 8  | set to -1 to set to 0
+	K                     int               // Default: 16 | set to -1 to set to 0
+	SineDeg               int               // Default: 30
+	DoubleAngle           int               // Default: 3  | set to -1 to not use double angle
+	ArcSineDeg            int               // Default: 0
+}
+
+func (p *ParametersLiteral) MarshalBinarySize() (dataLen int) {
```
ChristianMct, line 87: Same comment as for other literals.

ckks/advanced/homomorphic_encoding.go
```diff
@@ -20,48 +20,47 @@ const (
 // EncodingMatrix is a struct storing the factorized DFT matrix
 type EncodingMatrix struct {
```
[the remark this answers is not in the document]
Pro7ech: Done

ckks/advanced/homomorphic_mod.go
```diff
@@ -23,44 +23,38 @@ func cos2pi(x complex128) complex128 {
...
-	Sin  = SineType(0) // Standard Chebyshev approximation of (1/2pi) * sin(2pix)
-	Cos1 = SineType(1) // Special approximation (Han and Ki) of pow((1/2pi), 1/2^r) * cos(2pi(x-0.25)/2^r); this method requires a minimum degree of 2*(K-1).
-	Cos2 = SineType(2) // Standard Chebyshev approximation of pow((1/2pi), 1/2^r) * cos(2pi(x-0.25)/2^r)
+	CosDiscret    = SineType(0) // Special approximation (Han and Ki) of pow((1/2pi), 1/2^r) * cos(2pi(x-0.25)/2^r); this method requires a minimum degree of 2*(K-1).
+	SinContinuous = SineType(1) // Standard Chebyshev approximation of (1/2pi) * sin(2pix) on the full interval
+	CosContinuous = SineType(2) // Standard Chebyshev approximation of pow((1/2pi), 1/2^r) * cos(2pi(x-0.25)/2^r) on the full interval
 )
 
-// EvalModLiteral a struct for the parameters of the EvalMod step
+// EvalModLiteral a struct for the paramters of the EvalMod step
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/advanced/homomorphic_mod.go
```diff
@@ -23,44 +23,38 @@ func cos2pi(x complex128) complex128 {
...
+	LogScale        int      // Log2 of the scaling factor used during EvalMod
+	SineType        SineType // Chose betwenn [Sin(2*pi*x)] or [cos(2*pi*x/r) with double angle formula]
+	LogMessageRatio int      // Log2 of the tatio between Q0 and m, i.e. Q[0]/|m|
+	K               int      // K parameter (interpolation in the range -K to K)
+	SineDeg         int      // Degree of the interpolation
+	DoubleAngle     int      // Number of rescale and double angle formula (only applies for cos)
+	ArcSineDeg      int      // Degree of the Taylor arcsine composed with f(2*pi*x) (if zero then not used)
 }
 
 // EvalModPoly is a struct storing the EvalModLiteral with
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/bootstrapping/parameters.go
```diff
@@ -0,0 +1,635 @@
...
+//		dimension the scaling factors to be used: [level][scaling].
+//		For example: [][]int{{45}, {46}, {47}} means that the CoeffsToSlots step will use three levels, each
+//		with one prime. Primes are consumed in reverse order, so in this example the first matrix will use the
+//		prime of 47 bits, the second the prime of 46 bits, and so on.
+//		Non standard parameterization can include multiple scaling factors for a same prime, for example
+//		[][]int{{30}, {30, 30}} will use two levels for three matrices. The first two matrices will consume a prime
+//		of 30 + 30 bits, and have a scaling factor which prime^(1/2), and the third matrix will consume the second
+//		prime of 30 bits.
+//
+// S2CLogScale: the scaling factor and distribution of the moduli for the CoeffsToSlots (homomorphic encoding) step.
```
ChristianMct, line 45: SlotToCoeffs (homomorphic decoding)

ChristianMct (changes requested):

ckks/advanced/homomorphic_mod.go
```diff
@@ -135,26 +120,28 @@ func NewEvalModPolyFromLiteral(evm EvalModLiteral) EvalModPoly {
 		sqrt2pi = math.Pow(0.15915494309189535*qDiff, 1.0/scFac)
 	}
 
-	if evm.SineType == Sin {
+	if evm.SineType == SinContinuous {
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/bootstrapping/bootstrapping.go
Pro7ech: done

ckks/bootstrapping/dummy_bootstrapping.go
Pro7ech: done

ckks/bootstrapping/parameters.go
Pro7ech: done

ckks/bootstrapping/parameters.go
```diff
@@ -0,0 +1,635 @@
...
+	if paramsBootstrap.C2SLogScale == nil {
+		paramsBootstrap.C2SLogScale = [][]int{{56}, {56}, {56}, {56}}
+	}
+
+	if paramsBootstrap.S2CLogScale == nil {
+		paramsBootstrap.S2CLogScale = [][]int{{39}, {39}, {39}}
+	}
+
+	if paramsBootstrap.EvalModLogScale == 0 {
+		paramsBootstrap.EvalModLogScale = 60
```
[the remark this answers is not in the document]
Pro7ech: done

dckks/utils.go
```diff
@@ -17,18 +16,21 @@ import (
 // ok 		: a boolean flag, which is set to false if no such instance exist
 func GetMinimumLevelForBootstrapping(lambda int, scale rlwe.Scale, nParties int, moduli []uint64) (minLevel int, logBound uint, ok bool) {
```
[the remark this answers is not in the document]
Pro7ech: done

examples/rgsw/main.go
```diff
@@ -0,0 +1,127 @@
...
+// Function to evaluate
+func sign(x float64) float64 {
+	if x >= 0 {
+		return 1
+	}
+
+	return -1
+}
+
+func main() {
```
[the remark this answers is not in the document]
Pro7ech: done

ring/basis_extension.go
```diff
@@ -522,6 +514,51 @@ func reconstructRNS(index, x int, p [][]uint64, v *[8]uint64, y0, y1, y2, y3, y4
...
+	v[1] = uint64(vi[1])
+	v[2] = uint64(vi[2])
+	v[3] = uint64(vi[3])
+	v[4] = uint64(vi[4])
+	v[5] = uint64(vi[5])
+	v[6] = uint64(vi[6])
+	v[7] = uint64(vi[7])
+}
+
+func reconstructRNS(start, end, x int, p [][]uint64, v *[8]uint64, y0, y1, y2, y3, y4, y5, y6, y7 *[32]uint64, Q, QInv, QbMont []uint64) {
```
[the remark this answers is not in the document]
Pro7ech: yes

rlwe/bootstrapping.go
```diff
@@ -0,0 +1,25 @@
...
+	// For the collective bootstrapping it is given by the user-defined
+	// security parameters
+	StartingLevel() int
+
+	// EndLevel defines the level at which the bootstrapping ends.
+	// For the centralized bootstrapping this value is the maximum
+	// level minus the depth of the bootstrapping circuit.
+	// For the collective bootstrapping this value is usually the
+	// maximum level.
+	EndLevel() int
```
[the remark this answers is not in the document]
Pro7ech: done

rlwe/bootstrapping.go
```diff
@@ -0,0 +1,25 @@
...
+
+// Bootstrapper is a scheme independant generic interface to handle bootstrapping.
+type Bootstrapper interface {
+
+	// Bootstrapp defines a method that takes an empty interface as input and applies
+	// an in place scheme specific bootstrapping.
+	Bootstrap(ct interface{}) (err error)
+
+	// The depth of the bootstrapping circuit, can be zero.
+	Depth() int
```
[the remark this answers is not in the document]
Pro7ech: done

rlwe/bootstrapping.go
```diff
@@ -0,0 +1,25 @@
...
+
+	// StartingLevel defines the level at which the bootstrapping starts.
+	// For the centralized bootstrapping this value is usually zero.
+	// For the collective bootstrapping it is given by the user-defined
+	// security parameters
+	StartingLevel() int
```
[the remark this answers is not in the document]
Pro7ech: done

rlwe/params.go
```diff
@@ -119,7 +133,13 @@ func NewParameters(logn int, q, p []uint64, pow2Base, h int, sigma float64, ring
 		copy(params.pi, p)
 	}
 
-	return params, params.initRings()
+	if err = params.initRings(); err != nil {
+		if warning != nil {
```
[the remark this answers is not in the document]
Pro7ech: done

rlwe/params.go
```diff
@@ -155,17 +177,16 @@ func NewParametersFromLiteral(paramDef ParametersLiteral) (Parameters, error) {
...
 		var q, p []uint64
-		var err error
 		switch paramDef.RingType {
 		case ring.Standard:
 			q, p, err = GenModuli(paramDef.LogN, paramDef.LogQ, paramDef.LogP)
 		case ring.ConjugateInvariant:
 			q, p, err = GenModuli(paramDef.LogN+1, paramDef.LogQ, paramDef.LogP)
 		default:
-			return Parameters{}, fmt.Errorf("rlwe.NewParametersFromLiteral: invalid ring.Type, must be ring.ConjugateInvariant or ring.Standard")
+			return params, fmt.Errorf("rlwe.NewParametersFromLiteral: invalid ring.Type, must be ring.ConjugateInvariant or ring.Standard")
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/bootstrapping/parameters.go
Pro7ech: done

ckks/advanced/homomorphic_mod.go
```diff
@@ -104,16 +88,19 @@ func (evp *EvalModPoly) QDiff() float64 {
 	return evp.qDiff
 }
 
-// NewEvalModPolyFromLiteral generates an EvalModPoly from the EvalModLiteral.
-func NewEvalModPolyFromLiteral(evm EvalModLiteral) EvalModPoly {
+// NewEvalModPolyFromLiteral generates an EvalModPoly fromt the EvalModLiteral.
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/advanced/marshaler.go
Pro7ech: done

ckks/bootstrapping/bootstrapping.go
Pro7ech: done

rlwe/bootstrapping.go
```diff
@@ -0,0 +1,25 @@
+package rlwe
+
+// Bootstrapper is a scheme independant generic interface to handle bootstrapping.
+type Bootstrapper interface {
+
+	// Bootstrapp defines a method that takes an empty interface as input and applies
+	// an in place scheme specific bootstrapping.
+	Bootstrap(ct interface{}) (err error)
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/bootstrapping/dummy_bootstrapping.go
Pro7ech: done

ckks/bootstrapping/parameters_literal.go
```diff
@@ -0,0 +1,339 @@
...
+//
+// K: the range of the approximation interval, by default set to 16.
+//
+// SineDeg: the degree of the polynomial approximation of the modular reduction polynomial. By default set to 30.
+//
+// DoubleAngle: the number of double angle evaluation. By default set to 3.
+//
+// ArcSineDeg: the degree of the ArcSine Taylor polynomial, by default set to 0.
+type ParametersLiteral struct {
+	CoeffsToSlotsFactorizationDepthAndLogScales [][]int           // Default: [][]int{{56}, {56}, {56}, {56}}
```
ChristianMct, line 68: idea: 
- move this field (and the SlotToCoeffs) one could be in the `HomomorphicDFTMatrixLiteral` struct
- move the logic to generate the matrices parameters from `[][]int` in the homomorphic DFT constructor, the constructor checks if the level+moduli are specified, if not it revert to the normal construction from the other fields.
- have `ParametersLiteral` accept `HomomorphicDFTMatrixLiteral` as `CoeffToSlotParameter` field.

This way:
1. you have a user-friendlier constructor of encoding matrices in the advanced package
2. the bootstrapping is more parameterizable because it no longer hides the encoding matrices parameters
3. slightly better readability of the literals with something like:
```
ParametersLiteral{
   SlotsToCoeffsParams: HomomorphicDFTMatrixLiteral{FromFactorization: [][]int{{42}, {42}, {42}}} 
   ...
}
```

Not sure this is possible but it seems like the right thing to do.

ckks/bootstrapping/utils.go
```diff
@@ -0,0 +1,7 @@
+package bootstrapping
```
ChristianMct, line 1: Move to `utils` so we can use it elsewhere in the future ?

ckks/bootstrapping/utils.go
```diff
@@ -0,0 +1,7 @@
...
+
+// pointyInt re-implements point.Int() of https://github.com/openlyinc/pointy because this
+// helper pacakge requires Go v1.18 whereas Lattigo is backward compatible with up to Go v1.14.
+func pointyInt(x int) *int {
```
ChristianMct, line 5: This is quite a standard method, not sure this requires credits ;)

ckks/bootstrapping/simple_bootstrapper.go
```diff
@@ -0,0 +1,47 @@
...
+// uses the secret-key to decrypt and re-encrypt the bootstrapped ciphertext.
+type SimpleBootstrapper struct {
+	ckks.Parameters
+	ckks.Encoder
+	rlwe.Decryptor
+	rlwe.Encryptor
+	Counter int // records the number of bootstrapping
+}
+
+func NewSimpleBootstrapper(params ckks.Parameters, sk *rlwe.SecretKey) rlwe.Bootstrapper {
```
ChristianMct, line 18: This is not much more explicit than `Dummy`. What about `SkBootstrapper` or `SecretKeyBootstrapper` ?

CHANGELOG.md
```diff
@@ -13,15 +13,30 @@ All notable changes to this library are documented in this file.
...
 - CKKS: made `NttAndMontgomery` thread safe again!
 - CKKS: removed `Evaluator` methods `MultByGaussianInteger`, `MultByGaussianIntegerThenAdd`, `MultByi`, `MultByiNew`, `DivByi` and `DivByiNew`. These are now all handled by the methods `MultByConst[...]`.
 - CKKS: updated the behavior of `MultByConstAndAdd`.
 - CKKS: fixed the median statistics of `PrecisionStats`, that were off by one index. 
+- CKKS: renamed the `Parameters` field `DefaultScale` to `LogScale`, which now takes a value in log2.
+- CKKS: the `Parameters` field `LogSlots` now has a default value which is the maximum number of slots possible for the given parameters.
+- CKKS: variable `BSGSRatio` is now `LogBSGSRatio` and is given in log2.
+- CKKS/Bootstrapping: complete refactoring the bootstrapping parameters for better usability.
+- CKKS/Bootstrapping: upon bootstrapping, the method will check that the ciphertext scale is a power of two.
+- CKKS/Bootstrapping: added the iterative bootstrapping `META-BTS` of `Youngjin et al.`.
```
ChristianMct, line 27: [Youngjin et al.](<eprint link>) instead of  `Youngjin et al.`?

CHANGELOG.md
```diff
@@ -13,15 +13,30 @@ All notable changes to this library are documented in this file.
...
+- CKKS/Advanced: refactored names of structs and methods of the homomorphic encoding/decoding to better convey they purpose.
+- CKKS/Advanced: all fields of `EncodingMatrixLiteral` are now marshalled.
+- CKKS/Advanced: the `CoeffsToSlots` matrix is only scaled by an additional factor 1/2 if the `RepackImag2Real` field is set to true.
+- DCKKS: `GetMinimumLevelForBootstrapping` has been renamed to `GetMinimumLevelForRefresh`.
 - RLWE: added `CheckBinary` and `CheckUnary` to the `Evaluator` type. It performs pre-checks on operands of the `Evaluator` methods.
-- RLWE: added the `MaxLevelQ` and `MaxLevelP` methods to the `Parameters` type.
+- RLWE: added the methods `MaxLevelQ()` and `MaxLevelP` to the `Parameters` struct.
 - RLWE: added the method `NewCiphertextQP`.
-- RLWE: setting the Hamming weight of the secret or the standard deviation of the error to negative values will now instantiate these fields as zero values and return a warning.
+- RLWE: setting the Hamming weight of the secret or the standard deviation of the error to negative values will instantiate these fields as zero values and return a warning (as an error).
```
ChristianMct, line 36: Make it clearer that this is for the `NewParameter` only (the same behavior for `NewParameterFromLiteral` would be far too error-prone).

ckks/bootstrapping/parameters_literal.go
Pro7ech: I get the reasoning, but this is not possible. The `HomomorphicDFTMatrixLiteral` relies on a moduli chain that already exists. However this is not the case when creating the bootstrapping parameters. These additional parameters are needed there because the moduli chain for the bootstrapping circuit needs to be created. So they are not for the  `HomomorphicDFTMatrixLiteral` but for the creation of the moduli chain of the bootstrapping circuit.

The constructor for the `HomomorphicDFT` matrices in the `advanced` package is already very user friendly, it is difficult to do better. Here is an example:

```
// CoeffsToSlotsParameters homomorphic decoding parameters
var CoeffsToSlotsParameters = ckksAdvanced.HomomorphicDFTMatrixLiteral{
  Type:       ckksAdvanced.Encode,
  LogSlots:   LogSlots,
  LevelStart: 1,
  Levels:     []int{1},
}
```

Besides, if a user really want to give a custom linear transform, it is possible, since the fields in the `bootstrapping.Parameters` are public. So the option for more customization is there.

CHANGELOG.md
```diff
@@ -13,15 +13,30 @@ All notable changes to this library are documented in this file.
...
+- CKKS/Bootstrapping: upon bootstrapping, the method will check that the ciphertext scale is a power of two.
+- CKKS/Bootstrapping: added the iterative bootstrapping `META-BTS` of `Youngjin et al.`.
+- CKKS/Bootstrapping: added `SimpleBootstrapper` which provides a re-encryption using the secret key and complies to the `rlwe.Bootstrapper` interface.
+- CKKS/Advanced: refactored names of structs and methods of the homomorphic encoding/decoding to better convey they purpose.
+- CKKS/Advanced: all fields of `EncodingMatrixLiteral` are now marshalled.
+- CKKS/Advanced: the `CoeffsToSlots` matrix is only scaled by an additional factor 1/2 if the `RepackImag2Real` field is set to true.
+- DCKKS: `GetMinimumLevelForBootstrapping` has been renamed to `GetMinimumLevelForRefresh`.
 - RLWE: added `CheckBinary` and `CheckUnary` to the `Evaluator` type. It performs pre-checks on operands of the `Evaluator` methods.
-- RLWE: added the `MaxLevelQ` and `MaxLevelP` methods to the `Parameters` type.
+- RLWE: added the methods `MaxLevelQ()` and `MaxLevelP` to the `Parameters` struct.
```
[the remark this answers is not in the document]
Pro7ech: done

ckks/bootstrapping/parameters_literal.go
```diff
@@ -0,0 +1,339 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"math/bits"
+
+	"github.com/tuneinsight/lattigo/v4/ckks/advanced"
+)
+
+// ParametersLiteral is a struct to parameterize the bootstrapping parameters.
```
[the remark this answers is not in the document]
Pro7ech: Done

ckks/bootstrapping/utils.go
Pro7ech: done

ckks/bootstrapping/utils.go
Pro7ech: done

ckks/bootstrapping/simple_bootstrapper.go
Pro7ech: done

CHANGELOG.md
Pro7ech: done

CHANGELOG.md
Pro7ech: done

ckks/bootstrapping/parameters_literal.go
ChristianMct: Got it. Thanks!

CHANGELOG.md
ChristianMct: Remove the quotes so that the link is rendered correctly (see [current rendering](https://github.com/tuneinsight/lattigo/blob/dev_bootstrapping_2.0/CHANGELOG.md))

Pro7ech: Thank's, done

ChristianMct (approved):

ckks/bootstrapping/bootstrapper.go
```diff
@@ -232,7 +231,7 @@ func newBootstrapperBase(params ckks.Parameters, btpParams Parameters, btpKey Ev
 		bb.SlotsToCoeffsParameters.Scaling *= bb.params.DefaultScale().Float64() / (bb.evalModPoly.ScalingFactor().Float64() / bb.evalModPoly.MessageRatio())
 	}
 
-	bb.stcMatrices = advanced.NewHomomorphicEncodingMatrixFromLiteral(bb.SlotsToCoeffsParameters, encoder)
+	bb.stcMatrices = advanced.NewHomomorphicDFTMatrixFromLiteral(bb.SlotsToCoeffsParameters, encoder)
 
 	encoder = nil
```
romainbou, line 236: Is this for enhanced garbage collection or not needed ?

ckks/bootstrapping/bootstrapping.go
```diff
@@ -2,53 +2,86 @@
...
+		// 2^d * e
+		btp.MultByConst(tmp, 1<<16, tmp)
+
+		// 2^d * e + 2^(d-n) * e'
+		tmp = btp.bootstrap(tmp)
+
+		// 2^(d-n) * e + 2^(d-2n) * e'
+		btp.MultByConst(tmp, btp.params.QiFloat64(tmp.Level())/float64(uint64(1<<16)), tmp)
+		tmp.Scale = tmp.Scale.Mul(rlwe.NewScale(btp.params.Q()[tmp.Level()]))
+		btp.Rescale(tmp, btp.params.DefaultScale(), tmp)
```
romainbou, line 72: Should the return error of `Rescale()` be checked here?

romainbou (approved):

ckks/bootstrapping/bootstrapper.go
Pro7ech: Yes it is

ckks/bootstrapping/bootstrapping.go
Pro7ech: It doesn't need to, if this code is executed, then the appropriate prime is also present, one doesn't go without the other.

PR #303, opened 2022-11-30
fix issue#302

Pro7ech:
Added basic ECM factorization that is triggered when Pollard's rho factorization fails.

Also fixes a bug in Pollard's rho factorization.


Pro7ech:
Add marshalling of the root and prime factors in the params

ring/ecm.go
```diff
@@ -0,0 +1,193 @@
+package ring
```
ChristianMct, line 1: Factorization is only used in the `ring` package. Is there a specific reason we would want its interface to be public ? 

We'd better keep it private if it's not the case.

Pro7ech: In fact, I'd be more inclined to move the factorization to the `utils` package and have it imported by ring.

ChristianMct (approved):

PR #341, opened 2023-02-08
EvaluationKeys & Tests Refactoring

Pro7ech:
### Refactoring of Evaluation Keys

Currently, evaluation keys are defined by the struct `rlwe.SwitchingKey`. Although applying an evaluation key on a ciphertext does indeed serves the purpose of public re-encryption, user feedback has shown that this naming approach can lead to confusion and lacks an intuitive hierarchy among the different objects which are derived from the `rlwe.SwitchingKey` struct (and does not provide a comprehensive mapping with the literature). This is notably caused by the fact that public re-encryption is hardly ever used to change the access structure of a ciphertext, but as a mean to ensure ciphertext compactness and decryption correctness during the evaluation phase of a circuit.

To remedy to this issue, the struct `rlwe.SwitchingKey` has been renamed `rlwe.EvaluationKey`, and a comprehensive documentation on the generation, usage and all capabilities of this object, as well as code comments, have been added. Additionally, the method `SwitchKeys` has been renamed `ApplyEvaluationKey`.

The goal is to better convey that `rlwe.EvaluationKey` is a special type of public key that is used during the evaluation phase of a circuit, with different purposes (relinearization, rotations, automorphisms, etc...).

The struct `rlwe.RotationKeySet` has been removed. It is now replaced by the type `rlwe.GaloisKey`, which is a wrapper of `rlwe.EvaluationKey` and stores a single `rlwe.EvaluationKey` along with some meta data to help identify which automorphism it enables.

There is now a simple and much more intuitive hierarchy among `rlwe.EvaluationKeys`:

```
rlwe.EvaluationKey 
         |
         |-> rlwe.RelinearizationKey
         └-> rlwe.GaloisKey
```

To be able to manage multiple `rlwe.EvaluationKey`s, the interface `rlwe.EvaluationKeySetInterface` has been added. It enables the user to define it's own loading/saving/persistence policies for `rlwe.RelinearizationKey` and `rlwe.GaloisKey`s.

### Test Coverage and Consistency

Many tests and benchmarks located in the `bfv`, `bgv` and `ckks` packages were merely wrappers of methods of the package `rlwe`, and thus didn't serve any meaningful purpose. Additionally, many methods in the `rlwe` package where not fully tested or not tested at all.

To remedy to this issue, all tests and benchmarks in the `bfv`, `bgv` and `ckks` packages that were solely wrappers of a method located in `rlwe` have been removed. The test coverage of the `rlwe` package has been substantially increased by adding tests and expanding the range of parameters for which those methods are tested.

An example is the tests for automorphisms. These were tested in each of the schemes by checking that applying a specific automorphism on a ciphertext encrypting an encoded plaintext would result in a specific rotations in the plaintext decoded domain. Instead the `rlwe` package will now test that a homomorphic automorphism applies the correct permutation on the polynomial coefficients and that it produces the expected noise; and the schemes will only test that an automorphism on an encoded plaintext results in the correct rotation on the decoded plaintext.

Finally, all relevant methods of `rlwe` now have a benchmark function and the noise bounds for the tests in `rlwe` and `drwle` have been substantially improved.

##TODO:
- Check that marshalling tests are not done on empty structs.

## CHANGELOG
cf. `CHANGELOG.md`


Pro7ech:
Included in #379

PR #379, opened 2023-06-01
Release v5

ChristianMct:
https://github.com/tuneinsight/lattigo/releases/tag/v5.0.0

PR #383, opened 2023-07-03
Dev bootstrapping v4.3.0

Pro7ech:
## Done
- CKKS Bootstrapper:
  - Bootstrapping batches of ciphertexts of smaller dimension and/or with sparse packing with depth-less packing/unpacking
  - Bootstrapping for the Conjugate Invariant CKKS with optimal throughput

- CKKS Bootstrapping Circuit:
  - Improved implementation of [META-BTS](https://eprint.iacr.org/2022/1167) providing arbitrary precision bootstrapping from only one additional small prime
  - Generation of the bootstrapping parameters from pre-defined primes (previously only from LogQ)

- Example: 
  - MHE CKKS Bootstrapping Evaluation Keys Generation

## Todo
- Block composition bootstrapping


Pro7ech:
Included in #379

PR #451, opened 2024-04-14
[rlwe]: added .N() and .LogN() to rlwe.Element

Pro7ech:
This PR adds the methods `.N()` and `.LogN` to the `rlwe.Element`. 

This is a quality of life change that has two purposes:
- Enables the user to get an easy access to the value of ring degree or its base 2 logarithm directly from `rlwe.Ciphertext` or `rlwe.Plaintexts`. It is possible to already get these values by for example calling `bits.Len64(uint64(len(ct.Value[0].Coeffs[0]) - 1))`, however only experienced users will know that this is possible.
- Additionally, this improves code readability by replacing code such as `bits.Len64(uint64(len(ct.Value[0].Coeffs[0]) - 1))` by `ct.LogN()`.

Such information is for example needed when dealing with multiple parameter sets of different ring degree to be able to properly process an input ciphertext.

### DIFF
--- a/core/rlwe/element.go
+++ b/core/rlwe/element.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"io"
+	"math/bits"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/tuneinsight/lattigo/v5/ring"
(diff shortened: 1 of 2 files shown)


qantik:
Hi, can you give some context to this PR?

PR #473, opened 2024-05-14
Fix #457 `utils.MaxSlice` and `utils.MinSlice`

### LINKED ISSUE #457: Bug [utils]: some `rlwe` and `mhe` tests always pass due to a bug in `utils.MaxSlice` and `utils.MinSlice`
- `utils.MaxSlice` always return 0 if all values are negative
- `utils.MinSlice` always return 0 if all values are positive

This bug also cascades to another bug: the function `rlwe.NoiseGadgetCiphertext` will always return a zero noise because `BaseTwoDecompositionVectorSize` is zero and thus the last loop is never iterated. `rlwe.NoiseGadgetCiphertext` is used to check the correctness of the different evaluation keys in the package `rlwe` and `mhe`.

A user using this function, and its associated functions (e.g. `rlwe.NoiseRelinearizationKey`) will not be able to detect errors in its implementation, misleading him on the root of the issue.

romainbou:
Addressing #457

- Fix  `utils.MaxSlice` and `utils.MinSlice` returning 0
- Fix `EqualSlice` if slices have difference lengths
- Add tests for slices utils

### DIFF
--- a/core/rlwe/utils.go
+++ b/core/rlwe/utils.go
@@ -45,7 +45,7 @@ func NoiseGaloisKey(gk *GaloisKey, sk *SecretKey, params Parameters) float64 {
 	return NoiseEvaluationKey(&gk.EvaluationKey, skIn, skOut, params)
 }
 
-// NoiseGadgetCiphertext returns the log2 of the standard devaition of the noise of the input gadget ciphertext with respect to the given plaintext, secret-key and parameters.
+// NoiseGadgetCiphertext returns the log2 of the standard deviation of the noise of the input gadget ciphertext with respect to the given plaintext, secret-key and parameters.
 // The polynomial pt is expected to be in the NTT and Montgomery domain.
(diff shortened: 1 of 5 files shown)


Pro7ech:
The changes to `EqualSlice` make it not constant time anymore

Pro7ech:
> > `EqualSlice` is not constant time anymore after these changes
> 
> Ah yes, but before it would also panic early depending on the size of `b`.
> 
> It's essentially the implementation of the standard library. (available since Go 1.21, we could use it once we drop older Go version support) https://cs.opensource.google/go/go/+/refs/tags/go1.22.3:src/slices/slices.go;l=18
> 
> The only critical use is here:
> 
> https://github.com/tuneinsight/lattigo/blob/c031b14be1fb3697945709d7afbed264fa845442/mhe/keygen_evk.go#L131
> 
> Do you think it's really a concern?

To me it is not a concern since it is not used in critical code and if it is well documented (this is a cryptographic library). Since Go 1.21+ provides native support for this kind of utility, shouldn't it then be removed/replaced by it?

Issue #471, opened 2024-05-14
Bug[rlwe]: `Parameters.UnmarshalJSON`


Pro7ech:
Setting the JSON definition

```
"LogN":16,
"Q":[],
"P":[],
"LogQ": [60, 45, 45, 45, 45, 45, 45, 45, 45, 45, 45],
"LogP": [56, 55, 55, 55],
```
returns `rlwe.NewParametersFromLiteral: invalid parameter literal`

Might be linked to #458

Issue #466, opened 2024-05-09
"How is EncodingDomain set?"


yellow123Nike:
![image](https://github.com/tuneinsight/lattigo/assets/37546857/24995f70-d2d7-49fd-ab5c-fa210a1b33ab)
"How is EncodingDomain set?"

qantik:
Hi, the comment is a residual from an earlier Lattigo version. Whether a plaintext is in the slots or
coefficients domain is determined by the `IsBatched` field in the `PlaintextMetaData` object.

 - `IsBatched = true` -> slot encoding.
 - `IsBatched = false` -> coefficient encoding.

If the field is not set specifically by the user `IsBatched = true`.

Issue #465, opened 2024-05-08
Bug[rlwe]: `rlwe.Scale` serialization


Pro7ech:
The changes made in #456 now prevent from marshaling operands with a scale larger than $2^{100}-1$, however default scaling factors of up to $2^{128}$ are valid for the CKKS scheme.

Issue #459, opened 2024-05-02
Bug [rlwe]: `rlwe.Parameters.ReadFrom` fails (PR #456)


Pro7ech:
https://github.com/tuneinsight/lattigo/pull/456 added `WriteTo` and `ReadFrom` to the `rlwe.Parameters` without adding associated tests and it actually fails when calling `.ReadFrom`.
This can be easily reproduced by testing the serialization of `rlwe.Parameters` with the method `buffer.RequireSerializerCorrect`, which returns the error: `unexpected end of JSON input`.

This happens because `.BinarySize()` is called on empty parameters by `ReadFrom` to get the number of bytes to read. But this returns the wrong size because the serialization size of `rlwe.Parameters` is dependant on the content of the struct.

-> Do like other content-dependent structs (e.g. `ring.Poly`): add a 4/8 bytes prefix with the number of bytes to read

PR #474, opened 2024-05-17
Bugfixes May 2024 (Part 1)

qantik:
This PR bundles various bug fixes:

 - #457 (b326b02): Fix correctness errors in `utils.MaxSlice/MinSlice`. Prevent panic in `utils.EqualSlice` 
 - #459 (93154bd): Fix regression in `rlwe.Parameters` marshaller.
 - #466 (34a6325): Update reals tutorial explanation.
 - #458 / #471 (51451c4): Do not mix log/non-log primes in `rlwe.ParametersLiteral` initialisation.
 - #465 (454efab): Use decimal encoding to preserve `rlwe.Scale` to preserve binary size.
 - Bump minimum Go version to 1.21 with the use of `slices`
 
 - (25e88da): New bug reports policy.

### DIFF
--- a/core/rlwe/params.go
+++ b/core/rlwe/params.go
@@ -183,10 +183,11 @@ func NewParametersFromLiteral(paramDef ParametersLiteral) (params Parameters, er
 		paramDef.DefaultScale = s
 	}
 
+	// only allow either (Q or P) or (logQ or logP), do not mix log with non-log
 	switch {
-	case paramDef.Q != nil && paramDef.LogQ == nil:
+	case paramDef.Q != nil && paramDef.LogQ == nil && paramDef.LogP == nil:
 		return NewParameters(paramDef.LogN, paramDef.Q, paramDef.P, paramDef.Xs, paramDef.Xe, paramDef.RingType, paramDef.DefaultScale, paramDef.NTTFlag)
-	case paramDef.LogQ != nil && paramDef.Q == nil:
+	case paramDef.LogQ != nil && paramDef.Q == nil && paramDef.P == nil:
 		var q, p []uint64
 		switch paramDef.RingType {
 		case ring.Standard:
--- a/core/rlwe/params.go
+++ b/core/rlwe/params.go
@@ -1,6 +1,7 @@
 package rlwe
 
 import (
+	"bufio"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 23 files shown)


core/rlwe/params.go
```diff
@@ -181,10 +183,11 @@ func NewParametersFromLiteral(paramDef ParametersLiteral) (params Parameters, er
 		paramDef.DefaultScale = s
 	}
 
+	// only allow either (Q or P) or (logQ or logP), do not mix log with non-log
```
Pro7ech, line 186: There are use cases where one needs to specify `Q` to ensure that different parameter sets share the same ciphertext modulus (for example switching the ring degree), without having to give specific primes for `P` (i.e. let the library sample them). This prevents such use cases.

core/rlwe/scale.go
```diff
@@ -167,7 +167,7 @@ func (s Scale) Min(s1 Scale) (max Scale) {
 // BinarySize returns the serialized size of the object in bytes.
 // Each value is encoded with .Text('x', ceil(ScalePrecision / log2(10))).
 func (s Scale) BinarySize() int {
-	return 21 + (ScalePrecisionLog10+8)<<1 // 21 for JSON formatting and 2*(8 + ScalePrecisionLog10)
+	return 21 + (ScalePrecisionLog10+6)<<1 // 21 for JSON formatting and 2*(8 + ScalePrecisionLog10)
```
Pro7ech, line 170: Comments have not been updated to reflect the changes

core/rlwe/params.go
```diff
@@ -610,12 +613,15 @@ func (p Parameters) Equal(other *Parameters) (res bool) {
 // MarshalBinary returns a []byte representation of the parameter set.
 // This representation corresponds to the MarshalJSON representation.
 func (p Parameters) MarshalBinary() ([]byte, error) {
```
Pro7ech, line 615: Changes and additions to the parameters serialization are still not tested (and this previously caused a merge into main of code not working as expected)

core/rlwe/params.go
qantik: It's fixed.

core/rlwe/scale.go
qantik: Fixed.

core/rlwe/params.go
qantik: Fixed.

romainbou (approved):

PR #499, opened 2024-10-07
v6.1.0

lehugueni:
CHANGELOG:
- Update of `PrecisionStats` in `ckks/precision.go`:
  - The precision is now computed as the min/max/average/... of the log of the error (instead of the log of the min/max/average/... of the error).  
  - fields renamed (`MinPrecision` -> `MINLog2Prec`, `MaxPrecision` -> `MAXLog2Prec`, ...)
  - `rlwe.Scale` has a `.Log2()` method
- Update of `mod1.Parameters` fields (made public, some removed)
- Improvement of the relinearization key-generation protocol (reduce the degree of the shares)
- Serialisation of bootstrapping keys
- Lower noise incurred by `ModUp` (i.e. the modulus raise step in CKKS bootstrapping)
- Evaluation keys can be compressed (public element `a` can be generated from a seed)
- More doc formatting
- Fix various bugs:
  - `ShallowCopy` of the CKKS bootstrapping evaluator and BFV evaluator not deep enough.
  - PSI example failing
  - Incorrect reset of pointer in uniform sampler
  - Error when doing inverse NTT with small degree
  - Mod1Evaluator changes the input ciphertext


romainbou (approved):

core/rlwe/keygenerator.go
```diff
@@ -240,36 +258,56 @@ func (kgen KeyGenerator) GenEvaluationKey(skInput, skOutput *SecretKey, evk *Eva
...
 }
 
 func (kgen KeyGenerator) genEvaluationKey(skIn ring.Poly, skOut ringqp.Poly, evk *EvaluationKey) {
 
 	enc := kgen.WithKey(&SecretKey{Value: skOut})
 
+	// For a compressed evaluation key, a seed is created and stored in the EvaluationKey struct
+	// struct while an uncompressed key uses an ephemeral seed.
+	if evk.IsCompressed() {
+		evk.Seed = make([]byte, 32)
```
MGeorgie, line 282: what is the rng used for the seed?

core/rlwe/keygenerator.go
```diff
@@ -240,36 +258,56 @@ func (kgen KeyGenerator) GenEvaluationKey(skInput, skOutput *SecretKey, evk *Eva
...
+		if n, err := kgen.prng.Read(evk.Seed); n != 32 || err != nil {
+			panic(fmt.Errorf("unable to sample evaluation key seed"))
+		}
+
+		sampler, err := sampling.NewKeyedPRNG(evk.Seed)
```
MGeorgie, line 287: what  is the PRNG used here?

core/rlwe/metadata.go
```diff
@@ -146,20 +142,14 @@ func (m PlaintextMetaData) LogScale() float64 {
...
-func (m *CiphertextMetaData) Equal(other *CiphertextMetaData) (res bool) {
-	res = m.IsNTT == other.IsNTT
-	res = res && m.IsMontgomery == other.IsMontgomery
-	return
-}
-
 // BinarySize returns the size in bytes that the object once marshalled into a binary form.
 func (m PlaintextMetaData) BinarySize() int {
-	return 61 + m.Scale.BinarySize()
+	return 84 + m.Scale.BinarySize()
```
MGeorgie, line 152: what correspond 84?

MGeorgie (commented):
The PR seems good for me, I just have a few questions

core/rlwe/keygenerator.go
lehugueni: `NewKeyedPRNG` instantiates a XOF of Blake2b (https://pkg.go.dev/golang.org/x/crypto/blake2b#NewXOF) with `evk.Seed` as the key.

core/rlwe/keygenerator.go
lehugueni: The seed is obtained from reading 32 bytes out of a Blake2b XOF (the PRNG contained in `kgen.Encryptor`) keyed itself with a 64 bytes `key` sampled from `crypto/rand`

core/rlwe/ciphertext.go
```diff
@@ -43,6 +43,18 @@ func NewCiphertextRandom(prng sampling.PRNG, params ParameterProvider, degree, l
 	return
 }
 
+// Plaintext casts the target ciphertext into a plaintext type.
+// This method is allocation free.
+func (ct Ciphertext) Plaintext() *Plaintext {
```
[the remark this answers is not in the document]
Pro7ech: This doesn't change the modulus, only re-interprets the ciphertext (degree 1 poly in s) as a degree 0 poly in s. It can for example be used after key-switching a ciphertext to the key zero in a multi-party setting (instead of having to allocate a plaintext and manually setting it's internal value to turn first poly of the key-switches ciphertext, which would require a good knowledge of the low level elements library).

circuits/ckks/bootstrapping/evaluator.go
```diff
@@ -129,14 +129,30 @@ func NewEvaluator(btpParams Parameters, evk *EvaluationKeys) (eval *Evaluator, e
 // Evaluator can be used concurrently.
 func (eval Evaluator) ShallowCopy() *Evaluator {
 	heEvaluator := eval.Evaluator.ShallowCopy()
-	params := eval.BootstrappingParameters
+
+	paramsN1 := eval.ResidualParameters
```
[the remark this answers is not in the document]
lehugueni: `N1` in the variable name corresponds to the parameters outside of bootstrapping (i.e. the "residual parameters") and `N2` to the parameters used when bootstrapping (i.e. the "bootstrapping parameters").

core/rlwe/ciphertext.go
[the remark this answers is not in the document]
MGeorgie: Thanks for the information,  is the same method used for both CKKS and BGV/BFV? 
In CKKS, we agree that after encoding the plaintext and ciphertext space are the same. But that is not true for BFV and BGV.

core/rlwe/metadata.go
Pro7ech: The metadata are encode in a fixed-size JSON (each field has a fixed size value), 84 is the number of bytes of this JSON struct.

Pro7ech:
This PR modifies how the precision of CKKS is computed from the **arithmetic mean of the logs** to the **log of the geometric mean**. Both are good, but they mean different things. I believe geometric mean is more appropriate when looking at coefficients in the ring (more spikes) while the mean of the log is better when looking in the canonical embedding (more flat). Another benefit of the geometric mean is that it is much more efficient to compute when arbitrary arithmetic is involved (a substantial part of the test is computing those statistics).

In any case, the changelog and description should be updated accordingly to reflect this change and ideally, both stats should be made available.

core/rlwe/ciphertext.go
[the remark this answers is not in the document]
Pro7ech: Actually, it should be seen is a helper method for the following case: assume you have a ciphertext $(c_0, c_1)$, and you evaluate a decryption in place that modifies it to $(c_0 + sc_1, c_1)$, then by using `.Plaintext()` you wrap $(c0 + sc_1)$ in an `rlwe.Plaintext`, allowing to pass it to the `[scheme].Encoder`. Another case is, given a ciphertext $(0, 0)$, you could cast it as a plaintext to give it to an encoder and return $(encode(m), 0)$, and then add a mask on top of it. There are plenty of little cases that this helper makes easier to code, and it also helps it code readability.

Actually, the plaintext space of CKKS is not the same as the ciphertext space. The plaintext space of CKKS is $\mathbb{R}[X]/(X^N+1)$, so unbounded, while the ciphertext space is $\mathbb{Z}_{Q}[X]/(X^N+1)$, so bounded by $Q$. This is why $|m(X)|$ should always be smaller than $Q$.

Regarding the plaintext space of BGV/BFV, that depends how it is looked at: is it scaled by $t^{-1}\mod Q$ -> BFV or not -> BGV. Switching between both representations is trivial. By default, the encoder will scale plaintext by $t^{-1}\mod Q$ so it will assume that ciphertext encrypt such plaintext when decoding,

lehugueni:
> This PR modifies how the precision of CKKS is computed from the **arithmetic mean of the logs** 

I guess you meant log of the arithmetic means? 

I've edited the description and I'll update the changelog accordingly + create an issue to add support for both stats.

Pro7ech:
Yes, I meant the log of the arithmetic mean

MGeorgie (approved):
ok for me

PR #503, opened 2024-10-15
Security doc update

MGeorgie:
Update Security documentation


Pro7ech:
I have a few comments on the updated SECURITY.md. I usually would refrain to do that here, but this is an open PR on the topic and the fastest communication channel.

1. It says there are two approaches to mitigate the attack: the one currently implemented in Lattigo (specifying the bit-precision during the decoding step), and the one proposed by [Bossuat et al.](https://eprint.iacr.org/2024/853) (rounding off the noise). They actually are the same since they both round the noise of (should it be in the canonical embedding or in the ring, is another question). 

https://github.com/tuneinsight/lattigo/blob/661575ca7ec52f03b135ce676994e5e8b72188dd/schemes/ckks/encoder.go#L515

The contribution of Bossuat et al. is go give tight bounds on the noise to perform this rounding as efficiently as possible, limiting the loss in precision/efficiency.

2. Although rounding off the noise is sufficient to thwart the attack in the case where values would be disclosed by a honest client to third parties (which is the vast majority of cases), it unfortunately doesn't enable IND-CPA-D yet as it doesn't protects against an adversary that can craft arbitrary plaintexts. The reason is that since the error is mixed with the plaintext, the error cannot be cleanly rounded off as there will always exist messages where doing so will trigger a carry that will propagate and flip one ore bits above the rounding, enabling at the minimum an efficient distinguisher.

MGeorgie:
>     1. It says there are two approaches to mitigate the attack: the one currently implemented in Lattigo (specifying the bit-precision during the decoding step), and the one proposed by [Bossuat et al.](https://eprint.iacr.org/2024/853) (rounding off the noise). They actually are the same since they both round the noise of (should it be in the canonical embedding or in the ring, is another question).
> 
> 
> https://github.com/tuneinsight/lattigo/blob/661575ca7ec52f03b135ce676994e5e8b72188dd/schemes/ckks/encoder.go#L515
> 
> The contribution of Bossuat et al. is go give tight bounds on the noise to perform this rounding as efficiently as possible, limiting the loss in precision/efficiency.

Thanks for the clarification, I will update according.
> 
>     2. Although rounding off the noise is sufficient to thwart the attack in the case where values would be disclosed by a honest client to third parties (which is the vast majority of cases), it unfortunately doesn't enable IND-CPA-D yet as it doesn't protects against an adversary that can craft arbitrary plaintexts or ask for the evaluation of a circuit that would map a plaintext to carefully chosen values. The reason is that since the error is mixed with the message, it cannot be cleanly rounded off as there will always exist messages where doing so will trigger a carry that will propagate and flip one ore bits above the rounding, enabling at the minimum an efficient distinguisher.
If I understand well, the rounding is not enough and we need to add exponential amount of noise (noise flooding) to the decrypted value?

lehugueni (approved):
Approved by lattigo advisory committe.

PR #509, opened 2024-10-31
Fix serialization for compressed evaluation keys

lehugueni:
Currently, compressed evaluation keys cannot be serialized properly. 

This adds serialization methods to `EvaluationKey` that handle the `Seed` field (used when the key is compressed).


MGeorgie (approved):
good

core/rlwe/keys.go
```diff
@@ -417,6 +417,110 @@ func (evk EvaluationKey) Expand(params ParameterProvider, buffer *GadgetCipherte
...
+		var incInt int
+
+		if inc, err = evk.GadgetCiphertext.WriteTo(w); err != nil {
+			return n + inc, err
+		}
+
+		n += inc
+
+		if evk.IsCompressed() {
+			if incInt, err = w.Write(evk.Seed); err != nil {
```
Pro7ech, line 450: If you want a cleaner code where there isn't the need for converting between types of the incremental counter, there is the function `buffer.Write(w Writer, c []byte) (n int64, err error)`.

core/rlwe/keys.go
```diff
@@ -417,6 +417,110 @@ func (evk EvaluationKey) Expand(params ParameterProvider, buffer *GadgetCipherte
...
+				return n + int64(incInt), err
+			}
+
+			// Sanity check, should not happen unless the size of the seed has been modified in the code
+			if incInt != 32 {
```
Pro7ech, line 455: If this is mandatory, then I would change the seed type to `[32]byte` so that the size is natively enforced.

core/rlwe/keys.go
```diff
@@ -417,6 +417,110 @@ func (evk EvaluationKey) Expand(params ParameterProvider, buffer *GadgetCipherte
...
+				return n + int64(incInt), fmt.Errorf("writing compressed evaluation key: the size of the seed=%d != 32", incInt)
+			}
+
+			n += int64(incInt)
+		}
+
+		w.Flush()
```
Pro7ech, line 462: `.Flush()` can return an error

core/rlwe/keys.go
lehugueni: I have a small preference for leaving it as a slice as if the key is not compressed then `seed = nil`. ~This prevents using a nil seed when expanding.~ We should add a check when expanding that `seed != nil`.

Pro7ech: Then `*[32]byte` will should do it ;)

core/rlwe/keys.go
Pro7ech: By the way gosec should catch this kind of things.

lehugueni: Addressed in #511

PR #513, opened 2024-11-20
Fix inner sum in BGV

### LINKED ISSUE #510: Bug/Regression [BGV]: Wrong result from innerSum for params.N() elements
### What version of Lattigo are you using?
v6.1.0

### Does this issue persist with the latest release?
Yes

### What were you trying to do?
Let N = params.N().
Create a vector of length N, populate with random 1s and 0s.
Count the number of 1s by computing the inner sum of the encrypted vector with batch_size 1 and group size N

### What were you expecting to happen?
The result to match the number of 1s in the clear text bit vector.

### What actually happened?
The computed result was constantly slightly off from the expected result.
The answer was correct when
- Only filling the first N/2 slots of the plaintext and calling inner sum with group size N/2
OR
- Call to innerSum with group size N/2 and addition with RotateRowsNew

**Variant 1:**
clear text vector has length N with 1s all over.
inner sum is called with batch_size 1 and group size N
-> Incorrect result

**Variant 2:**
clear text vector has length N with 1s all over.
inner sum is called with batch_size 1 and group size N >> 1
Addition with RotateRowsNew
-> Correct result

**Variant 3:** 
clear text vector has length N >> 1 with 1s all over.  
inner sum is called with batch_size 1 and group size N >> 1
-> **Correct** result

"I think the library should return an error when asking innersum with N or handle this case internaly in the BGV side (because the backend of innersum is in the RLWE package, which is agnostic of the scheme and packing, hence why it doesn't return an error). The BGV evaluator should re-implement this method by calling the RLWE innersum but then add this last step if n = N." -Jean-Philippe Bossuat

### Reproducibility
Minimal example with variant 1.
Variant 2 is achieved by removing comments around RotateRowsNew
Variant 3 is achieved by replacing all 'params.N()' by 'params.N() >> 1'

```
package main

import (
	"fmt"
	"github.com/tuneinsight/lattigo/v6/core/rlwe"
	"github.com/tuneinsight/lattigo/v6/schemes/bgv"
	"math/rand/v2"
)

func innerSumTest() {
	// Setup copied from examples
	var err error
	var params bgv.Parameters

	if params, err = bgv.NewParametersFromLiteral(bgv.ExampleParameters128BitLogN14LogQP438); err != nil {
		panic(err)
	}

	kgen := rlwe.NewKeyGenerator(params)

	sk := kgen.GenSecretKeyNew()
	ecd := bgv.NewEncoder(params)

	enc := rlwe.NewEncryptor(params, sk)
	dec := rlwe.NewDecryptor(params, sk)

	decrypt := func(ct *rlwe.Ciphertext) []uint64 {
		pt := bgv.NewPlaintext(params, params.MaxLevel())
		dec.Decrypt(ct, pt)

		decoded := make([]uint64, params.N())
		if err = ecd.Decode(pt, decoded); err != nil {
			panic(err)
		}
		return decoded
	}

	rlk := kgen.GenRelinearizationKeyNew(sk)
	gal := params.GaloisElementsForInnerSum(1, params.N())

	evk := rlwe.NewMemEvaluationKeySet(rlk)

	galk := kgen.GenGaloisKeysNew(gal, sk)

	eval := bgv.NewEvaluator(params, evk)
	evalks := rlwe.NewMemEvaluationKeySet(rlk, galk...)
	eval = eval.WithKey(evalks)

	//Init inputs
	values1 := make([]uint64, params.N())

	sum := 0
	for i := range values1 {
		if rand.Int()%3 == 0 {
			values1[i] = 1
			sum += 1
		}
	}
	fmt.Printf("Input: %v...\n", values1[:100])

	//Encode + encrypt
	pt1 := bgv.NewPlaintext(params, params.MaxLevel())

	if err = ecd.Encode(values1, pt1); err != nil {
		panic(err)
	}

	var ct *rlwe.Ciphertext
	if ct, err = enc.EncryptNew(pt1); err != nil {
		panic(err)
	}

	//Eval inner sum
	tmp := bgv.NewCiphertext(params, 1, params.MaxLevel())
	if err = eval.InnerSum(ct, 1, params.N(), tmp); err != nil {
		panic(err)
	}
/*
	var tmpRot *rlwe.Ciphertext
	if tmpRot, err = eval.RotateRowsNew(tmp); err != nil {
		panic(err)
	}

	if err = eval.Add(tmp, tmpRot, tmp); err != nil {
		panic(err)
	}
*/

	//Verify output
	decRes := decrypt(tmp)
	fmt.Printf("Expected: %d\n", sum)
	fmt.Printf("Computed: %d\n", decRes[0])
}
```

Edited: RotateRowsNew does compute the correct result

lehugueni:
Fix #510 

### Issue 
`bgv.Evaluator.InnerSum` was not working properly when `n*batchSize > N/2` as `InnerSum` was acting separately on the "rows" of the underlying plaintext. 

### New behaviour
Let `l := n*batchSize`. 
- If `l <= n/2`: `bgv.Evaluator.InnerSum` acts as before (i.e. the inner sum is computed on both rows "independently").
- If `l = n`: the inner sum is computed over the `n` slots via a rotation of the rows and an addition as showcased in #510 
- Otherwise: return an error

### DIFF
--- a/schemes/bgv/evaluator.go
+++ b/schemes/bgv/evaluator.go
@@ -1505,6 +1505,43 @@ func (eval Evaluator) RotateHoistedLazyNew(level int, rotations []int, op0 *rlwe
 	return
 }
 
+// InnerSum computes the inner sum of the underlying slots (see [rlwe.Evaluator.InnerSum]).
+// NB: in the slot encoding of BGV/BFV, the underlying N slots are arranged as 2 rows of N/2 slots.
+// If n*batchSize < N/2, InnerSum computes the [rlwe.Evaluator.InnerSum] of each row separately.
+// If n*batchSize = N, InnerSum computes the [rlwe.Evaluator.InnerSum] on the concatenation of both rows.
+// NOTE: In this case, InnerSum performs an addition and a [Evaluator.RotateRowsNew] on top
+// Otherwise, InnerSum returns an error.
+func (eval Evaluator) InnerSum(ctIn *rlwe.Ciphertext, batchSize, n int, opOut *rlwe.Ciphertext) (err error) {
+	N := eval.parameters.N()
+	halfN := N >> 1
+	l := n * batchSize
+
+	if l > halfN {
+		if l != N {
+			return fmt.Errorf("innersum: n*batchSize=%d > N/2=%d and n*batchSize != N=%d", l, halfN, N)
+		}
+
+		if err = eval.Evaluator.InnerSum(ctIn, batchSize, n/2, opOut); err != nil {
+			return
+		}
+
+		var ctRot *rlwe.Ciphertext
+		ctRot, err = eval.RotateRowsNew(opOut)
+		if err != nil {
+			return
+		}
+
+		if err = eval.Add(opOut, ctRot, opOut); err != nil {
+			return
+		}
+
+		return
+	}
+
+	err = eval.Evaluator.InnerSum(ctIn, batchSize, n, opOut)
+	return
+}
+
 // MatchScalesAndLevel updates the both input ciphertexts to ensures that their scale matches.
 // To do so it computes t0 * a = opOut * b such that:
 //   - ct0.Scale * a = opOut.Scale: make the scales match.
--- a/schemes/bgv/params.go
+++ b/schemes/bgv/params.go
@@ -257,7 +257,7 @@ func (p Parameters) GaloisElementForRowRotation() uint64 {
 // InnerSum operation with parameters batch and n.
 func (p Parameters) GaloisElementsForInnerSum(batch, n int) (galEls []uint64) {
 	galEls = rlwe.GaloisElementsForInnerSum(p, batch, n)
-	if n > p.N()>>1 {
+	if n*batch > p.N()>>1 {
 		galEls = append(galEls, p.GaloisElementForRowRotation())
 	}
 	return


schemes/bgv/evaluator.go
```diff
@@ -1505,6 +1505,44 @@ func (eval Evaluator) RotateHoistedLazyNew(level int, rotations []int, op0 *rlwe
 	return
 }
 
+// InnerSum computes the inner sum of the underlying slots (see [rlwe.Evaluator.InnerSum]).
+// NB: in the slot encoding of BGV/BFV, the underlying N slots are arranged as 2 rows of N/2 slots.
+// If n*batchSize is a multiple of N, InnerSum computes the [rlwe.Evaluator.InnerSum] on the N slots.
+// NOTE: In this case, InnerSum performs an addition and a [Evaluator.RotateRowsNew] on top.
+// Otherwise, InnerSum computes the [rlwe.Evaluator.InnerSum] of each row separately.
+func (eval Evaluator) InnerSum(ctIn *rlwe.Ciphertext, batchSize, n int, opOut *rlwe.Ciphertext) (err error) {
```
Pro7ech, line 1513: Overall I think it would be clearer to simply add a flag to add the rows together, this would be more natural of the plaintext space. This will also remove the need for a complicated description of how the method behaves.

schemes/bgv/params.go
```diff
@@ -257,7 +257,7 @@ func (p Parameters) GaloisElementForRowRotation() uint64 {
 // InnerSum operation with parameters batch and n.
 func (p Parameters) GaloisElementsForInnerSum(batch, n int) (galEls []uint64) {
 	galEls = rlwe.GaloisElementsForInnerSum(p, batch, n)
-	if n > p.N()>>1 {
+	if n*batch%p.MaxSlots() == 0 {
```
Pro7ech, line 260: if `n` or `batch = 0` then it will add this key, but these are invalid inputs.

schemes/bgv/evaluator.go
```diff
@@ -1505,6 +1505,44 @@ func (eval Evaluator) RotateHoistedLazyNew(level int, rotations []int, op0 *rlwe
...
+	N := eval.parameters.MaxSlots()
+	l := n * batchSize
+
+	if l%N == 0 {
+		if n == 1 {
+			if ctIn != opOut {
```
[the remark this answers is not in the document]
lehugueni: Ah yes indeed, I think this check could be removed in a few other places then.

schemes/bgv/evaluator.go
```diff
@@ -1505,6 +1505,44 @@ func (eval Evaluator) RotateHoistedLazyNew(level int, rotations []int, op0 *rlwe
...
 }
 
+// InnerSum computes the inner sum of the underlying slots (see [rlwe.Evaluator.InnerSum]).
+// NB: in the slot encoding of BGV/BFV, the underlying N slots are arranged as 2 rows of N/2 slots.
+// If n*batchSize is a multiple of N, InnerSum computes the [rlwe.Evaluator.InnerSum] on the N slots.
+// NOTE: In this case, InnerSum performs an addition and a [Evaluator.RotateRowsNew] on top.
+// Otherwise, InnerSum computes the [rlwe.Evaluator.InnerSum] of each row separately.
+func (eval Evaluator) InnerSum(ctIn *rlwe.Ciphertext, batchSize, n int, opOut *rlwe.Ciphertext) (err error) {
+	N := eval.parameters.MaxSlots()
+	l := n * batchSize
```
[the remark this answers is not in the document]
lehugueni: I think we should properly define what are valid parameters for this function (and the underlying one in `rlwe`). 
For instance `InnerSum` "works" with `n*batchSize > N`, i.e. it wraps around and keep adding, but it's not optimal performance-wise and I don't see why one would need to do that. 

A list of *invalid* parameters could be: 
 * `n = 0` or `batchSize = 0`
 * `n*batchSize > N` as you suggested
 * `N > n*batchSize > N/2` only for `bgv.Evaluator.InnerSum`

schemes/bgv/evaluator.go
lehugueni: I'm not for adding a flag as I think most users expect `InnerSum` to do an inner sum on the whole `N` slots, without worrying about the internal representation. 
For now, the function behaves like that if `n * batchSize` is a power of 2, which I believe covers most flagship use cases of the function. 
If `n * batchSize` is not a power of 2, the result will not be the inner sum on the whole `N` slots anyway, regardless of the flag (adding the rows will not give the correct result).

schemes/bgv/evaluator.go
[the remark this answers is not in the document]
Pro7ech: Yes indeed, or the domain of valid parameters (if it is shorter to list).

schemes/bgv/evaluator.go
lehugueni: Yes I understand, but my point is that most users are not aware of the structure of the underlying plaintext space as the issue linked to this PR highlights. 
In the current fix, if `n*batchSize` is a power of 2 (i.e. divides the #of slots), the result is correct whether the user is aware of the 2D structure of the plaintext space and expects an inner sum on each row, or they expect an inner sum on the cleartext vector. This covers most common use cases and the savvy user can easily aggregate rows if they want.  

If it is not a power of 2, the result will not be a "textbook" inner sum anyway as the sum on the last batches will wrap around. Then adding the possibility to aggregate the results of these inner sum variants makes this function even more different than what I'd expect an inner sum function to be. This would cover only niche use cases and it seems quite thin to warrant an API change in my opinion.

Pro7ech: I think users should be made aware of the 2D structure of the plaintext, else we will have more issues about it as it will not behave as how they would expect it in other operations, such as rotations. 

Besides this is not an API change since this API is not yet present in the BGV package, so this is the opportunity to make it right. 

The goal isn't to cover nich use cases, but to be correct, as this is the API directly interfacing with the scheme. If higher APIs are built on top of it, people are free to do anything they want to abstract the hypercube structure.

lehugueni: This is an API change in the sense that legacy code with calls of the form `bgv.InnerSum(old_parameters)` will fail if we add another parameter as there is no method overloading in Go. 

Also, I don't see how allowing aggregation makes the inner sum more correct. If anything it makes it more different than a textbook innersum and we don't offer aggregation on other operations (`Add`, `Mul`, etc.).

I agree that the correct way would be to let the user do only the inner sum on the rows as before. I just think allowing for the special case `n*batchSize=N` is fine because it clearly marks the intent of the user, it answers a documented need that must quite common and simplify the life of users. 

As a side note, I'm all for adding a `Rotate` method besides the `RotateRows` and `RotateColumns` ones in the long term, so so we'd have the basic operations `Add, Mul, Rotate` that can act on the cleartext vector. But that's another debate :P.

Pro7ech: I'm fine with it working for `n*batchsize=N`, but I disagree that the plaintext algebra should be abstracted already at the scheme level. My experience is that the lowest API should be as flexible as possible, because there is nowhere to fall back to if it happen to not be. 

The fact that the `bgv.Evaluator` already has an `InnerSum` API through the `rlwe.Evaluator` is more a result of how Go propagates methods when structs are wrapper into structs rather than a deliberate choice to make the `rlwe.Evaluator.InnerSum` available through the `bgv.Evaluator`. In fact I even think it should not be called `InnerSum` at the `rlwe` package level, because this package has no concept of batching.

This is something that we can discuss during the next meeting.

Pro7ech (changes requested):
I would rename `rlwe.Evaluator.PartialTrace` to `rlwe.Evaluator.PartialTracesSum` or something similar because here we have an accumulator `accQP` that sums HW(n) partial traces. It's only when n is a power of two (where HW(n) = 1) and offset divides n that we have a true partial Trace (in the sens that full trace is from 0 to LogN and partial means from a to b where 0<= a < b <= logN). In other words, .Trace is a special case of `PartialTracesSum`.

CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 # Changelog
 All notable changes to this library are documented in this file.
 
+## [6.x.x] - 16.12.2024
+- Refactoring of the InnerSum methods: 
+  - `rlwe.Evaluator.InnerSum` has been replaced by `rlwe.Evaluator.PartialTrace`
```
MGeorgie, line 6: add desription what PartialTrace / PartialTraceSum computes

MGeorgie (approved):
I'm  happy with this solution

lehugueni:
> I would rename `rlwe.Evaluator.PartialTrace` to `rlwe.Evaluator.PartialTracesSum` or something similar because here we have an accumulator `accQP` that sums HW(n) partial traces. It's only when n is a power of two (where HW(n) = 1) and offset divides n that we have a true partial Trace (in the sens that full trace is from 0 to LogN and partial means from a to b where 0<= a < b <= logN). In other words, .Trace is a special case of `PartialTracesSum`.

Yes I agree, it is a partial trace in the sense that it is a sum with a subset of the automorphisms, but you're right that `Trace` can also be partial in that sense. I've changed to `PartialTracesSum` for now, we can rename it a later time if someone finds a better suited name.

Pro7ech (approved):

PR #533, opened 2025-01-13
Range of NTTLazy

### LINKED ISSUE #524: Bug[`ring`]: NTTLazy returns values outside of the range [0, 2q-1]
To be clear, this is not a correctness issue and it can also be treated as a documentation issue and cleared by updating the documentation to the range [0, 4q-2]. However the convention for the suffix `lazy` in Lattigo usually means the output values are expected to be in the range [0, 2q-1]. Predictable range is especially important when doing lazy additions, to know how many additions can be done without modular reduction before values wrap around u64.

```Go
package main

import(
	"fmt"
	"github.com/tuneinsight/lattigo/v6/ring"
)

func main(){
	n := 16
	r, err := ring.NewRing(n, []uint64{65537})

	if err != nil{
		panic(err)
	}

	p := r.NewPoly()

	for i := range n{
		p.Coeffs[0][i] = uint64(i)
	}

	r.NTTLazy(p, p)

	fmt.Println(p)
}
```

lehugueni:
Fix #524 

NTTLazy doc now states that the output is in the range `[0, 6q-2]`. 

I preferred a change of documentation over a post-NTT lazy reduction as: 

- a lazy reduction adds an overhead. It is best to have one less efficient function that preserves the range (`NTT`) and another one that privileges efficiency (`NTTLazy`) and gives flexibility to the user to reduce further if needed. 
- there are already several methods that have the suffix 'Lazy' but don't output values in the `[0, 2q-1]` range. Lazy should be more indicative of whether the output is in `[0, q-1]`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     - name: Setup Go
       uses: actions/setup-go@v5
       with:
-        go-version: '1.23.5'
+        go-version: '1.23.6'
 
     - uses: actions/cache@v4
       with:
@@ -34,7 +34,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.23.5', '1.22.11', '1.21.13']
+        go: ['1.23.6', '1.22.11', '1.21.13']
 
     steps:
       - uses: actions/checkout@v4


MGeorgie (approved):
Ok to change the doc and keep the large range, just for my understanding what is the reason to have output in the [0, 6q-2]? Should we also point to the function that allow to normalize in [0, 2q-1]?

Pro7ech (approved):

PR #537, opened 2025-01-23
Thread-safety: replace buffers thread-safe pools

lehugueni:
This PR replaces the temporary buffers stored in objects with thread-safe pools that can be used to obtain temporary objects. 

- [x] Added the `structs.BufferPool` interface and `structs.SyncPool`, `structs.BuffFromUintPool` that implement it.
- [x] Modify `ring.Ring` so that it stores a `structs.SyncPool[*[]uint64]`
- [x] Add methods that build pools of `ring.Poly`, `ringqp.Poly` and `rlwe.Ciphertext` from `structs.SyncPool[*[]uint64]`  
- [ ] Replace buffers with these pools in all non-thread-safe structures of FHE: 
  - [x] `rlwe.Evaluator`
  - [x] `rlwe.Decryptor`
  - [x] `ring.BasisExtender`
  - [x] `ckks.Evaluator`
  - [x] `ckks.Encoder`
  - [x] `bgv.Evaluator`
  - [x] `bgv.Encoder`
  - [ ] `sampler.Gaussian`
  - [ ] `sampler.Uniform`
  - [ ] `rlwe.Encryptor`
  - [ ] `rlwe.KeyGenerator`
- [ ] Replace buffers with pools in all non-thread-safe structures of MHE
- [ ] Add tests for correctness of concurrent executions


ring/sampler_gaussian.go
```diff
@@ -74,9 +72,12 @@ func (g *GaussianSampler) read(pol Poly, f func(a, b, c uint64) uint64) {
 
 	r := g.baseRing
 
+	randomBufferN := make([]byte, 1024)
```
ChristianMct, line 75: Could be an array.

utils/sampling/prng.go
```diff
@@ -2,49 +2,53 @@ package sampling
...
+func (prng *ThreadSafePRNG) Read(sum []byte) (n int, err error) {
+	return rand.Read(sum)
+}
+
+// KeyedPRNG is a structure storing the parameters used to securely and *deterministically* generate shared
 // sequences of random bytes among different parties using the hash function blake2b. Backward sequence
 // security (given the digest i, compute the digest i-1) is ensured by default, however forward sequence
 // security (given the digest i, compute the digest i+1) is only ensured if the KeyedPRNG is keyed.
+// WARNING: KeyedPRNG should NOT be called by multiple threads. It does not make sense to do so as the resulting
+// sequence will not be deterministic for a given key. For a PRNG securely seeded with a private keyuse [ThreadSafePRNG].
```
ChristianMct, line 34: Unclear then why it is protected by a mutex. Either the type or the documentation has to be changed.

lehugueni: I updated the comment. In short, this should not be used concurrently as it doesn't make sense (the result won't be deterministic). However, if someone does (inadvertently or on purpose for an obscure reason), the resulting sequences will be independent and "safe" to use.

MGeorgie (approved):
ok for me

ring/ringqp/pool.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"fmt"
+
+	"github.com/tuneinsight/lattigo/v6/ring"
+	"github.com/tuneinsight/lattigo/v6/utils/structs"
+)
+
+// Pool represents a pool of polys that can be used (concurrently) to instantiate temporary polynomials in RingQP.
+type Pool struct {
```
ChristianMct, line 11: Same as for ring level pools.

ring/pool.go
```diff
@@ -0,0 +1,61 @@
+package ring
+
+import (
+	"fmt"
+
+	"github.com/tuneinsight/lattigo/v6/utils/structs"
+)
+
+// Pool represents a pool of polys that can be used (concurrently) to instantiate temporary polynomials.
+type Pool struct {
```
ChristianMct, line 10: This type name is a bit vague (a pool is a generic concept). `PolyPool` or `BufferPool` or `MemPool` ? Other suggestions welcome.

core/rlwe/pool.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"fmt"
+
+	"github.com/tuneinsight/lattigo/v6/ring"
+	"github.com/tuneinsight/lattigo/v6/ring/ringqp"
+	"github.com/tuneinsight/lattigo/v6/utils/structs"
+)
+
+// Pool represents a pool of different objects (plaintexts, ciphertexts, polys) that can be used to instantiate temporary buffers.
+type Pool struct {
```
ChristianMct, line 12: Same as for ring level Pools.

ring/pool.go
ChristianMct: I guess `PolyPool` makes sense at this level but not at higher level where there are other types of pooled values like ciphertexts.

We might also want to reflect that on some field names (but less a priority as most are private anyway).

utils/sampling/prng.go
ChristianMct: The new phrasing makes it much clear. Thanks!

core/rlwe/encryptor.go
```diff
@@ -313,7 +297,7 @@ func (enc Encryptor) encryptZeroPkNoP(pk *PublicKey, ct Element[ring.Poly]) (err
 
 	ringQ := enc.params.RingQ().AtLevel(levelQ)
 
-	buffQ0 := enc.buffQP[0].Q
+	buffQ0 := enc.params.ringQ.NewPoly()
```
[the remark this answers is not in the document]
ChristianMct: So with the new internal buffers, should we make a pass on those places where we can save allocations ?

ChristianMct (approved):

PR #558, opened 2025-07-28
Fix `DivRoundByLastModulus`

### LINKED ISSUE #531: [`ring`]: `DivRoundByLastModulus` modifies the left operand
https://github.com/tuneinsight/lattigo/blob/1d70f2ced3e57b59d7b1534053b76cdd09003cb8/ring/scaling.go#L113

This method should not write on `p0`.

```Go
func (r Ring) DivRoundByLastModulus(p0, p1 Poly) {

	level := r.level

	// Center by (p-1)/2
	pHalf := (r.SubRings[level].Modulus - 1) >> 1

	r.SubRings[level].AddScalar(p0.Coeffs[level], pHalf, p0.Coeffs[level])

	for i, s := range r.SubRings[:level] {
	        // p0 is modified
		s.AddScalarLazyThenNegTwoModulusLazy(p0.Coeffs[i], s.Modulus-BRedAdd(pHalf, s.Modulus, s.BRedConstant), p0.Coeffs[i])
		s.AddLazyThenMulScalarMontgomery(p0.Coeffs[level], p0.Coeffs[i], r.RescaleConstants[level-1][i], p1.Coeffs[i])
	}
}
```

lehugueni:
Fixes #531

### DIFF
--- a/ring/scaling.go
+++ b/ring/scaling.go
@@ -3,16 +3,21 @@ package ring
 // DivFloorByLastModulusNTT divides (floored) the polynomial by its last modulus.
 // The input must be in the NTT domain.
 // Output poly level must be equal or one less than input level.
-func (r Ring) DivFloorByLastModulusNTT(p0, buff, p1 Poly) {
+func (r Ring) DivFloorByLastModulusNTT(p0, p1 Poly) {
 
 	level := r.level
 
-	r.SubRings[level].INTTLazy(p0.Coeffs[level], buff.Coeffs[0])
+	buff0 := r.pool.GetBuffUintArray()
+	defer r.pool.RecycleBuffUintArray(buff0)
+	buff1 := r.pool.GetBuffUintArray()
+	defer r.pool.RecycleBuffUintArray(buff1)
+
+	r.SubRings[level].INTTLazy(p0.Coeffs[level], *buff0)
 
 	for i, s := range r.SubRings[:level] {
-		s.NTTLazy(buff.Coeffs[0], buff.Coeffs[1])
+		s.NTTLazy(*buff0, *buff1)
 		// (-x[i] + x[-1]) * -InvQ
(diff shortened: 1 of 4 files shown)


MGeorgie (approved):
oki

ring/scaling.go
```diff
@@ -109,19 +109,19 @@ func (r Ring) DivRoundByLastModulusNTT(p0, buff, p1 Poly) {
 }
 
 // DivRoundByLastModulus divides (rounded) the polynomial by its last modulus. The input must be in the NTT domain.
-// Output poly level must be equal or one less than input level.
-func (r Ring) DivRoundByLastModulus(p0, p1 Poly) {
+// Output poly level must be equal or one less than input level. Buffer poly level must be larger or equal to input level.
```
Pro7ech, line 112: The minimum size can be reduced to 2 since it requires 1 limb for the INTT(q[level]) and 1 limb for NTT(q[i])

ring/scaling.go
```diff
@@ -143,7 +143,7 @@ func (r Ring) DivRoundByLastModulusManyNTT(nbRescales int, p0, buff, p1 Poly) {
 
 			rCpy.INTT(p0, buff)
 			for i := 0; i < nbRescales; i++ {
-				rCpy.DivRoundByLastModulus(buff, buff)
+				rCpy.DivRoundByLastModulus(buff, buff, buff)
```
Pro7ech, line 146: Doc should include the minimum size of buff

ring/scaling.go
lehugueni: If buff at level 0 and 1 is used, then one could not use `DivRoundByLastModulus` with the buffer and the output poly as the same variable as it's done now.  

I realized that a ring has already a pool so I just removed the buffers there to use the pool to be consistent.

Pro7ech (approved):

PR #562, opened 2025-09-02
Fix: multiple matrices can be applied for same prime in homomorphic dft

### LINKED ISSUE #561: Bug: homomorphic DFT correctness error when decomposition level has multiple matrices for a single prime
This bug is at the origin of https://github.com/tuneinsight/lattigo/discussions/560

## Cause

https://github.com/tuneinsight/lattigo/blob/c930849f741895887789a3e2df068042b58be9d6/circuits/ckks/dft/dft.go#L343

```Go
// dft evaluates a series of [lintrans.LinearTransformation] sequentially on the ctIn and stores the result in opOut.
func (eval *Evaluator) dft(ctIn *rlwe.Ciphertext, matrices []ltcommon.LinearTransformation, opOut *rlwe.Ciphertext) (err error) {

	inputLogSlots := ctIn.LogDimensions

	// Sequentially multiplies w with the provided dft matrices.
	if err = eval.LTEvaluator.EvaluateSequential(ctIn, matrices, opOut); err != nil {
		return
	}

	// Encoding matrices are a special case of `fractal` linear transform
	// that doesn't change the underlying plaintext polynomial Y = X^{N/n}
	// of the input ciphertext.
	opOut.LogDimensions = inputLogSlots

	return
}
```

Does not take into account that multiples matrices can be evaluated for a single prime. 

## Fix
Algorithm shouldn't call EvaluateSequential, but sequentially evaluate each matrice and only rescale after all the matrices of a same level have been evaluated (this can be derived from the `Level` field of `MatrixLiteral`).

lehugueni:
Fix #561 as described in the issue. The bug was raised in #560. 

Added a test checking correctness of the bootstrapping for all default parameters. Not run by default, but only with `-allparams` flag.

### DIFF
--- a/circuits/ckks/bootstrapping/default_parameters.go
+++ b/circuits/ckks/bootstrapping/default_parameters.go
@@ -181,13 +181,13 @@ var (
 	N15QP880H16384H32 = defaultParametersLiteral{
 		ckks.ParametersLiteral{
 			LogN:            15,
-			LogNthRoot:      17,
 			LogQ:            []int{40, 31, 31, 31, 31},
 			LogP:            []int{56, 56},
 			Xs:              ring.Ternary{H: 16384},
 			LogDefaultScale: 31,
 		},
 		ParametersLiteral{
+			LogN: utils.Pointy(15),
 			SlotsToCoeffsFactorizationDepthAndLogScales: [][]int{{30, 30}},
 			CoeffsToSlotsFactorizationDepthAndLogScales: [][]int{{52}, {52}},
 			EvalModLogScale: utils.Pointy(55),


MGeorgie (approved):
oki

circuits/ckks/bootstrapping/default_parameters.go
```diff
@@ -181,6 +181,7 @@ var (
 	N15QP880H16384H32 = defaultParametersLiteral{
 		ckks.ParametersLiteral{
 			LogN:            15,
+			LogNthRoot:      17,
```
Pro7ech, line 184: Why is 4N = 1 mod Q necessary here?

lehugueni: Because this only sets the value `LogN` for the residual parameters, but the `LogN'` of the bootstrapping parameters will be the default one (i.e. `16`) and we need `Q = 1 mod 2N'`.

Pro7ech: But the residual paramters for this set are expected to have LogN=15 too, so I dont' see why LogNthRoot should be hardcoded to 17 for default parameters (it's ok to manually set it though, but I think that for default it shouldn't). Is it because the tests fix LogN to 16? Why not derive it from the bootstrapping parameters?

lehugueni: Yes, the residual parameters will have LogN=15 but the bootstrapping params will have LogN=16 so we need to generate the primes accordingly. It all follows from the implementation of `NewParametersFromLiteral` and its doc: https://github.com/tuneinsight/lattigo/blob/5fc229b046cbba8ccbc00dd528af50e9de2307d5/circuits/ckks/bootstrapping/parameters.go#L46

Pro7ech: But is a set for Logn=15

Pro7ech: Ok, I understand the confusion. These default parameters are from the papers and are expecting LogN residual parameters to be equal to LogN bootstrapping parameters. Of course you can accommodate the parameters for other ring degree, but these are not the "default" ones anymore (refering to the papers) and will not reproduce exactly the moduli chain used for the experiments.

PR #572, opened 2026-01-29
Enabling `ring.Poly` argument to `bgv.Encoder.Encode`

ChristianMct:
I stumbled over these two limitations with the current BGV encoder:

1. It fails silently on unsupported input types when the flag `pt.IsBatched=false`.
2. It only allows for `IntegerSlice` parameters, while it could be useful to many applications to encrypt `ring.Poly` directly. It is possible with the current API but is quite tricky and undocumented.

This PR address these issues by defining the missing cases and encoder semantic. It is currently done as a minimal change, but a more sequential implementation of the Encoding workflow could be considered. The PR also adds a `ckks` test case for issue 1, even though `ckks.Encoder` currently behaves as expected in this case.


schemes/bgv/test_utils.go
```diff
@@ -87,8 +87,8 @@ func VerifyTestVectors(params Parameters, encoder *Encoder, decryptor *rlwe.Decr
 		t.Error("invalid unsupported test object type")
 	}
 
-	fmt.Println("have", values[:10])
-	fmt.Println("want", want[:10])
+	// fmt.Println("have", values[:10])
+	// fmt.Println("want", want[:10])
```
lehugueni, line 91:
```suggestion
```

lehugueni (approved):
thanks!

MGeorgie (approved):
good

PR #576, opened 2026-06-02
Secret key encryption with PRNG

lehugueni:
As discussed in the last committee meeting, this (re-)adds support for secret key encryption with the public element sampled from a PRNG. 
I.e. generate a ciphertext $$(a*sk + e, a)$$ where $$a$$ is generated using a PRNG passed as input.


MGeorgie (approved):
ok for me

core/rlwe/encryptor.go
```diff
@@ -122,6 +122,38 @@ func newTestEncryptorWithKeyedPRNG(params ParameterProvider, key EncryptionKey,
...
 }
 
+// EncryptSKWithPRNG encrypts the input plaintext using the stored secret key and writes the result on ct = (a*sk + e + pt, a).
+// The public element a is sampled uniformly using the PRNG passed as input.
+// The method currently accepts only [*Ciphertext] as ct.
+// If a [Plaintext] is given, then the output [Ciphertext] [MetaData] will match the [Plaintext] [MetaData].
+// The method returns an error if the ct has an unsupported type or if no secret key is stored in the [Encryptor].
+//
+// WARNING: This method *MUST NOT* be called twice with the same PRNG and secret key.
+// Reusing the same PRNG/key pair would produce correlated RLWE samples of the form a*s+e, a*s+e', ... which completely breaks security.
```
ChristianMct, line 132: Saying "completely" isn't very precise and could be omitted.

core/rlwe/encryptor.go
```diff
@@ -122,6 +122,38 @@ func newTestEncryptorWithKeyedPRNG(params ParameterProvider, key EncryptionKey,
 	return enc
 }
 
+// EncryptSKWithPRNG encrypts the input plaintext using the stored secret key and writes the result on ct = (a*sk + e + pt, a).
+// The public element a is sampled uniformly using the PRNG passed as input.
+// The method currently accepts only [*Ciphertext] as ct.
+// If a [Plaintext] is given, then the output [Ciphertext] [MetaData] will match the [Plaintext] [MetaData].
```
ChristianMct, line 128: Might be more general than this PR but the doc actually doesn't specify that `pt = nil` is equivalent to a zero plaintext. It is also unspecified what happens with the metadata in this case.

Issue #579, opened 2026-08-31
`bignum.MonomialEval()` omits the highest-degree coefficient and panics on constant polynomials


CCYJ1014:
## Summary

At Lattigo commit `5dbffbdea05394de2ca3a432ed5318aa832e3f40`, the public helper `utils/bignum.MonomialEval()` is documented as:

```go
// MonomialEval evaluates y = sum x^i * poly[i].
```

However, the implementation starts Horner evaluation from `poly[len(poly)-2]` instead of `poly[len(poly)-1]`:

```go
func MonomialEval(x *big.Float, poly []*big.Float) (y *big.Float) {
	n := len(poly) - 1
	y = new(big.Float).Set(poly[n-1])
	for i := n - 2; i >= 0; i-- {
		y.Mul(y, x)
		y.Add(y, poly[i])
	}
	return
}
```

This has two observable effects for ordinary valid monomial polynomials:

- a constant polynomial, such as `[5]`, panics with `index out of range [-1]`;
- a polynomial with degree at least one silently ignores its highest-degree coefficient, for example evaluating `[1, 2]` at `x = 3` as `1` instead of `1 + 2*x = 7`.

This is a local numeric correctness bug in a public helper.

## Environment

- Repository: <https://github.com/tuneinsight/lattigo>
- Tested commit: `5dbffbdea05394de2ca3a432ed5318aa832e3f40`
- Module: `github.com/tuneinsight/lattigo/v6`
- `go.mod`: `go 1.25.0`
- Toolchain used for the probe: `go version go1.25.0 linux/amd64`

The probe module below uses a local `replace` directive. The tested source is therefore the local checkout at commit `5dbffbdea05394de2ca3a432ed5318aa832e3f40`;
the `v6.0.0` requirement only satisfies the Go module declaration.

## Minimal Reproduction

`go.mod`:

```text
module lattigo-bignum-monomial-eval-probe

go 1.25.0

require github.com/tuneinsight/lattigo/v6 v6.0.0

replace github.com/tuneinsight/lattigo/v6 => /path/to/lattigo
```

`main.go`:

```go
package main

import (
	"fmt"
	"math/big"
	"os"

	"github.com/tuneinsight/lattigo/v6/utils/bignum"
)

func bf(v float64) *big.Float {
	return new(big.Float).SetPrec(128).SetFloat64(v)
}

func printEval(name string, x *big.Float, coeffs ...*big.Float) {
[...]

PR #580, opened 2026-09-05
fix(bignum): MonomialEval drops highest-degree coefficient and panics on constants

dualfroz:
## Problem

**Base:** `tuneinsight/lattigo:main` (commit 5dbffbd)
**Head:** `dualfroz:dualfroz/fix-monomialeval-highest-coeff`
**Issue:** #579 (free - no assignee, no linked PR when picked).

---

**Title:**

fix(bignum): MonomialEval drops the highest-degree coefficient and panics on constants

**Body:**

`utils/bignum.MonomialEval` is documented as `y = sum x^i * poly[i]`, but the
Horner evaluation started one coefficient too low:

```go
n := len(poly) - 1
y = new(big.Float).Set(poly[n-1]) // should be poly[n]
for i := n - 2; i >= 0; i-- {       // should be n-1
```

Consequences on valid input:
- the highest-degree coefficient is silently ignored, e.g. `1 + 2x` at `x = 3`
  evaluates to `1` instead of `7`;
- a constant polynomial such as `[5]` indexes `poly[-1]` and panics with
  `index out of range [-1]`.

This starts Horner from `poly[n]` and iterates from `n-1`, which is the standard
evaluation and matches the documented contract. The only in-tree caller,
`Remez.eval` for the `Monomial` basis, benefits directly.

Added `TestMonomialEval` covering a linear and a quadratic polynomial and the
constant case (there was no existing test for this helper).

---


lehugueni:
Hi, thanks for the effort but this does not fit into our [external contribution policy](https://github.com/tuneinsight/lattigo#pull-requests).

